PR #29, opened 2015-06-23
Cap

acozine:
Cap deploy enabled, production database gem added, JavaScript runtime enabled.

### DIFF
--- a/config/deploy.rb
+++ b/config/deploy.rb
@@ -1,0 +1,48 @@
+# config valid only for current version of Capistrano
+lock '3.4.0'
+
+set :application, 'my_app_name'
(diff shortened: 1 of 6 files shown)


config/deploy.rb
```diff
@@ -0,0 +1,41 @@
+# config valid only for current version of Capistrano
+lock '3.4.0'
+
+set :application, 'sufia'
```
hackmastera: In general can we use 'chf-sufia' or something that otherwise makes it clear semantically that we are installing an app based on the sufia engine (as opposed to sufia itself, which can't be run as an app because it is a gem / engine)?

PR #164, opened 2016-02-22
Add 'stereographs' to genre list

catlu:


hackmastera:
@sanfordd I thought you might like to do this one?

PR #171, opened 2016-05-18
Changes to Rights statements and URIs, fixes #170

### LINKED ISSUE #170: Update RightsStatement.org URIs to actual addresses

catlu:
Adding in the correct URIs to RightsStatements.org now that the website is up (Anna and I had guessed at the links before they existed).


hackmastera:
I'm not sure but I this one may necessitate data changes on deployment.

hackmastera:
I'm not sure but I think this one may necessitate data changes on deployment.

hackmastera:
i'm nearly certain that the uri is stored as the value of the field, so if these are changed then the values that people selected previously may not show up in the UI any longer, causing them to become angry that data was lost (although technically it hasn't been lost) and leading them to re-add fields they've already saved. Then the field will be on there twice, once with the old URI and once with the new one. Then there will be a mix of records with the field on there just once (with the old URI) vs. twice (with both URIs) which will make cleanup slightly more complicated.

catlu:
yikes, haha. @sanfordd, let's wait on this?

sanfordd:
Given that complexity, I'm definitely willing to wait on that.

MDiMeo:
Thanks for catching this @hackmastera! We will certainly wait till you're back to push this out.

hackmastera:
to be handled in the upgrade to sufia 7; see 67ebf0b

PR #183, opened 2016-07-27
Added 'Postage stamps' to Genre dropdown

catlu:


hackmastera:
merged into branches manually.

PR #360, opened 2017-03-01
dynamically look up fits_path

jrochkind:
On my dev machine, fits is just called 'fits'

@hackmastera what do you think of this solution?


hackmastera:
I like it! and it works for me.

PR #369, opened 2017-03-06
Transform rights uri to label for facet display, fixes #353

### LINKED ISSUE #353: Rights facet should display as text, not uri

hackmastera:


app/controllers/catalog_controller.rb
```diff
@@ -58,7 +58,7 @@ def self.modified_field
     config.add_facet_field solr_name("medium", :facetable), label: "Medium", limit: 5
     config.add_facet_field solr_name("language", :facetable), label: "Language", limit: 5
     config.add_facet_field solr_name("division", :facetable), label: "Division", limit: 5
-    config.add_facet_field solr_name("rights", :facetable), label: "Rights", limit: 5
+    config.add_facet_field solr_name("rights", :facetable), helper_method: :license_label, label: "Rights", limit: 5
```
jrochkind: neat, good to know you can do this, I didn't. The helper method is just used for display of facet values, I guess? Neat. Does the `CurationConcerns::LicenseService` require a non-cached lookup to fedora? Cause it could slow things down a lot if it did. Let's look it up and see....  

There is actually so much abstraction involved that I'm not even certain. https://github.com/projecthydra/curation_concerns/blob/master/app/services/curation_concerns/qa_select_service.rb#L25

What do you think about putting in some caching in the #license_label helper, just in case?  I suspect these labels change very infrequently, cache for an hour?  Could look like:

~~~
def license_label(id)
    Rails.cache.fetch("chf-sufia/license_label/#{id}", expires_in: 1.hour) do
       CurationConcerns::LicenseService.new.label(id)
    end
end
~~~

The Rails cache by default uses a 'memory store', just local RAM -- which would probably be fine, although we might want to give it a max size, not sure what the default is. (http://guides.rubyonrails.org/caching_with_rails.html#cache-stores).  **BUT**, since we have a Redis dependency anyway, using Redis as the `Rails.cache` backend would be a very common/standard thing to do.

hackmastera: Ha, great instinct to worry about retrieval speed from Fedora. In this case it's grabbing the value straight from a [config file](https://github.com/chemheritage/chf-sufia/blob/master/config/authorities/licenses.yml) (via the [questioning_authority](https://github.com/projecthydra-labs/questioning_authority) gem). So I don't think we need caching. That's interesting to see, though, since I've never done that sort of thing before.

jrochkind: Awesome, sounds good, will merge. 

Yeah, `Rails.cache` is super useful for caching arbitrary expensive things. Hopefully `questioning_authority` caches the file read, instead of doing it each time you make the call, but if not that's a problem for the gem... oh, and it looks like it's actually a ruby class rather than a textual config file, so that'd happen automatically? Anyway, enough in the weeds on this one!

We should prob go ahead and switch Rails.cache to use redis at some point, just as a good practice. But not part of this PR.  (The `Rails.cache` backend is also used for fragment caching if we ever use it)

PR #372, opened 2017-03-07
dev/test setup

jrochkind:
Easier workflows for dev running and testing using wrappers, with docs
Travis setup


.solr_wrapper
```diff
@@ -2,8 +2,12 @@
 # version: 6.0.0
 # port: 8983
 instance_dir: tmp/solr-development
+  <% # in subdir of tmp to make it easier to tell travis to cache,
```
hackmastera: this is good to cache, since solr imposes daily download limits.

lib/tasks/dev.rake
```diff
@@ -0,0 +1,63 @@
+require 'rspec/core'
+require 'rspec/core/rake_task'
+require 'solr_wrapper'
+require 'fcrepo_wrapper'
+require 'active_fedora/rake_support'
+
+module DevRakeUtil
+  def load_config(service, environment, defaults)
```
hackmastera: Maybe comment what this is overriding? Something in active_fedora or one / both of the wrappers?

jrochkind: oops, this is unnecceary and didn't need to be committed. Was a false path. Thanks!

jrochkind:
Okay, this is ready! And travis is green!

spec/factories/generic_works.rb
```diff
@@ -77,7 +77,10 @@
       end
     end
 
-
-
+    # a public work with complete metadata and a real image! Slow.
```
hackmastera: If you'd like it to be faster, you could do an option without any files at all. there are plenty of metadata and relationship-y things you can mess around with using works alone, and our implementation doesn't require a work to have a file attached to it (whereas default sufia does).

hackmastera:
this looks awesome; i only noticed that [one commit message](5c29d1c3dbcec0dbb39e6c6ed67691d4ee98f52e) says 'OpenDataControllerSpec' but should be 'OpacDataControllerSpec'

Maybe let's talk about it briefly in the morning. i'm super excited to have travis working!

PR #376, opened 2017-03-08
remove config/role_map.yml

jrochkind:
I _think_ this file is not actually used by our app, and can be
deleted. Not totally sure, but as a start we'll see if tests still
pass.

Deleting dead code makes the project cleaner and more comprehensible,
without false leads when trying to figure out what's going on.


hackmastera:
@jrochkind I prefer to rebase and then force-push the feature branch, rather than merging master in -- it gives a cleaner commit history. Is there a reason you like to do it this way instead?

jrochkind:
I tend to prefer to avoid rebases unless truly necessary, since it's a 'dangerous' thing to do and has gotten me in trouble before, caused me lots of time recovering from weirdness.  So I just merge master when that works fine.

PR #382, opened 2017-03-09
Change from single to double quotations to properly encode control ch…

hackmastera:
…aracters

If you use single quotes it goes in a a literal \r\n and then renders as such in the UI.


jrochkind:
oops, thanks.

PR #384, opened 2017-03-12
Add desired fields to 'all fields' search, fixes #351

### LINKED ISSUE #351: List of fields that should be searched with keyword 'all fields' search

hackmastera:
This PR includes 2 other cleanup-type commits for my own convenience but let me know if you prefer them split out.

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -16,6 +16,46 @@ class CatalogController < ApplicationController
     solr_name('system_modified', :stored_sortable, type: :date)
   end
 
+  def self.chf_search_fields
+    [
+      solr_name("title", :stored_searchable),
+      solr_name("depositor"),
+      solr_name("description", :stored_searchable),
+      solr_name("subject", :stored_searchable),
+      solr_name("language", :stored_searchable),
+      solr_name("identifier", :stored_searchable),
+      solr_name("related_url", :stored_searchable),
+      solr_name("artist", :stored_searchable),
+      solr_name("author", :stored_searchable),
+      solr_name("addressee", :stored_searchable),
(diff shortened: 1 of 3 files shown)


jrochkind:
@hackmastera It's interesting that the Hydra/Sufia way to do this seems to be index every property in their own Solr field, and then list all the solr fields in the solr configuration to search.  In the past, I would have done this different -- if there were different properties which should all be searchable with the same level of 'relevant boosting', I'd aggregate them in the _same_ Solr field; that's more how Solr is intended to be used I think, but this alternate Hydra way should work too, sure. 

but I'm gonna ask on Hydra Slack what the context for this is, cause I'm curious. 

Also gonna spend some more time with this PR as an entry point to understanding what's going on. 

(Note, I'm not asking for changes, I'll merge it, just discusisng).

hackmastera:
@jrochkind good point. thanks!

jrochkind:
I think this approach is going to be a lot less performant for Solr, to have to include dozens of fields in the search instead of a handful. But that may not matter at the scale we are at.

hackmastera:
We'll have to index them all anyway, but it would be easy to create an 'keyword search' solr copy field instead and then define that as the target for 'all fields'.

jrochkind:
Why do we have to index them all seperately anyway?

Here's the slack discussion. https://project-hydra.slack.com/archives/dev/p1489420198226279

hackmastera:
The reason we index them separately is so they can be displayed on a record page without hitting fedora.

hackmastera:
right; re: the slack discussion, we looked at the local indexer a bit when we were in the gerstacker room. it's here https://github.com/chemheritage/chf-sufia/blob/master/app/indexers/generic_work_indexer.rb

PR #386, opened 2017-03-13
rake task to fix String dates, fixes #322

### LINKED ISSUE #322: date created / modified not displaying for migrated works
They were copied over as strings and cannot be indexed that way.

run through all works, check the class of each date and parse them if needed
```
work.date_uploaded.class == String
work.date_uploaded = DateTime.parse(work.date_uploaded)
work.date_modified = DateTime.parse(work.date_modified)
```
brought to our attention by PSU
https://github.com/psu-stewardship/scholarsphere/compare/i%23574?expand=1#diff-c5f78d950658d0c37e52fc25fc0938baR22

jrochkind:
RAILS_ENV=production bundle exec rake chf:data_fix:fix_work_dates

Still in progress on staging, seems to be taking hours, not sure if it
really should be expected to?

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -13,6 +13,9 @@ gem 'curation_concerns', github: 'projecthydra/curation_concerns', branch: '1-6-
 # pull in fix to "add another" label
 gem 'hydra-editor', github: 'projecthydra/hydra-editor', ref: 'c1e9d298'
 
+# used in some rake tasks, some of which we may want ot run on production.
+gem 'ruby-progressbar', '~> 1.0'
+
 # Needed to fix reindex_everything
 # But don't go to 11 without being ready for rdf2
 gem 'active-fedora', '~>10.3'
--- a/lib/chf/data_fixes/work_dates.rb
+++ b/lib/chf/data_fixes/work_dates.rb
@@ -1,0 +1,26 @@
+module CHF
+  module DataFixes
+    class WorkDates
+      attr_reader :work
+
+      def initialize(work)
+        @work = work
+      end
+
+      # possibly makes changes to work (without saving), returns true
+      # if changes were made, so caller can do expensive save.
+      def change
+        updated = false
+        if work.date_uploaded != nil && work.date_uploaded.is_a?(String)
+          work.date_uploaded = DateTime.parse(work.date_uploaded)
+          updated = true
+        end
+        if work.date_modified != nil && work.date_modified.is_a?(String)
+          work.date_modified = DateTime.parse(work.date_modified)
+          updated = true
+        end
+        return updated
+      end
+    end
+  end
+end
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -868,6 +868,7 @@ DEPENDENCIES
   rsolr (~> 1.0)
   rspec-activemodel-mocks
   rspec-rails (~> 3.1)
+  ruby-progressbar (~> 1.0)
   sass-rails (~> 5.0)
   sdoc (~> 0.4.0)
   solr_wrapper (>= 0.3)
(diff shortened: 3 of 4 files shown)


jrochkind:
Going REALLY slow in staging, not sure why it seems to get slower and slower the more records it goes through. @hackmastera , have you encountered this before, how long do you _expect_ something to take that needs to read and save a good proportion of the collection? Is this just how hydra is?

hackmastera:
tasks where you save are definitely slower than tasks where you don't save anything. Depending on the number of things you need to save (and in this case it's probably close to 2000) I would expect it to take hours. Actually I think this has been related to our solr performance issues? @sanfordd?

I wouldn't expect this to be getting slower as it goes. Fedora gets slower as anything gets bigger: more objects in general or more objects in any given container. But this is just modifying existing objects, so... :\

jrochkind:
okay, I have trouble letting it run to completion cause the ssh connection gets broken. Maybe there is a better way to run this than ssh'ing to staging server, do you do something else?

But I think this is done/ready/working. Want to try it yourself on staging or something?

jrochkind:
Okay, this has run to completion in staging. So you can check out staging and make sure it hasn't done anything awful to the db or anything, and has fixed what needed fixing?

hackmastera:
In general when I run jobs on staging or production I use screen or tmux.

The simplest way to use it is to just invoke screen:
`$ screen`

Then go about your business; to detatch do `ctrl-a d` and to reattach type `screen -r`.

If you lose your connection your screen session should still be running; just ssh back onto the machine and type `screen -r`

It's possible to make additional windows, etc, and if you're actually going to use it as a work environment I would do that sort of thing. But to just run a job you don't really need to know anything else.

Note if you use tmux I've added my tmux dotfiles to most of the servers for the ubuntu user and the meta command (usually `ctrl-b`) is `ctrl-space`

lib/chf/data_fixes/work_dates.rb
```diff
@@ -0,0 +1,26 @@
...
+      attr_reader :work
+
+      def initialize(work)
+        @work = work
+      end
+
+      # possibly makes changes to work (without saving), returns true
+      # if changes were made, so caller can do expensive save.
+      def change
+        updated = false
```
hackmastera: There is a `.changed?` that I was just messing around with but it doesn't appear to be universally useful:

```
[21] pry(main)> w.visibility
=> "open"
[22] pry(main)> w.visibility = "restricted"
=> "restricted"
[23] pry(main)> w.changed?
=> false
[24] pry(main)> w.title = ["hello"]
=> ["hello"]
[25] pry(main)> w.changed?
=> true
```

Gemfile
```diff
@@ -13,6 +13,9 @@ gem 'curation_concerns', github: 'projecthydra/curation_concerns', branch: '1-6-
 # pull in fix to "add another" label
 gem 'hydra-editor', github: 'projecthydra/hydra-editor', ref: 'c1e9d298'
 
+# used in some rake tasks, some of which we may want ot run on production.
+gem 'ruby-progressbar', '~> 1.0'
```
hackmastera: love this idea

lib/chf/data_fixes/work_dates.rb
jrochkind: interesting idea, but especially if we don't know if it's reliable, i'm happy with how i've done it. this is reporting if THIS operation made a change specifically, too. Although if we could rely on `changed?` it wouldn't have to. But I think this is good enough, the hackiness is localized.

hackmastera: Yeah there is a separate `visibility_changed?` as well. So maybe that is the only one that isn't caught? idk; I agree it seems better not to rely on it.

PR #387, opened 2017-03-13
Truncate description in search results

### LINKED ISSUE #331: Truncate description in search results
After how many characters? Okay to truncate mid-word? add ellipsis.

jrochkind:
Closes #331

This is a totally insane way to do this, but I think might be
the bet we've got?

When/if we overhaul our front-end, I would anticipate possibly doing
the front-end templates in a different way, cause this architecture
is wacky. We'll see.

### DIFF
--- a/app/presenters/chf_index_presenter.rb
+++ b/app/presenters/chf_index_presenter.rb
@@ -1,0 +1,26 @@
+class ChfIndexPresenter < Blacklight::IndexPresenter
+  class_attribute :description_max_chars
+  self.description_max_chars = 10
+
+  attr_reader :description_field
+
+  def initialize(*args)
+    super
+    # This is an attempt to not hard-code in that our field is called
+    # `description_tesim`, so that doesn't break if it's changed, we
+    # just look for the "description" field. Not really sure if this is
+    # really less fragile, or if there was a better place to get this.
+    @description_field = configuration.index_fields.values.
+      find { |conf| conf.itemprop == "description" }.
+      field
+  end
+
+  def field_value(field, *other_args)
+    case field
+    when description_field
+      view_context.truncate(super, length: description_max_chars, separator: /\s/)
+    else
+      super
+    end
+  end
+end
--- a/app/presenters/chf_index_presenter.rb
+++ b/app/presenters/chf_index_presenter.rb
@@ -1,6 +1,6 @@
 class ChfIndexPresenter < Blacklight::IndexPresenter
   class_attribute :description_max_chars
-  self.description_max_chars = 10
+  self.description_max_chars = 400
 
   attr_reader :description_field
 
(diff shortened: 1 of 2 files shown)


app/presenters/chf_index_presenter.rb
```diff
@@ -0,0 +1,26 @@
...
+  attr_reader :description_field
+
+  def initialize(*args)
+    super
+    # This is an attempt to not hard-code in that our field is called
+    # `description_tesim`, so that doesn't break if it's changed, we
+    # just look for the "description" field. Not really sure if this is
+    # really less fragile, or if there was a better place to get this.
+    @description_field = configuration.index_fields.values.
+      find { |conf| conf.itemprop == "description" }.
```
hackmastera: does this work if a field is indexed more than one way? (e.g. `index.as :stored_searchable, :facetable`) I don't think description is, though.

app/presenters/chf_index_presenter.rb
```diff
@@ -0,0 +1,26 @@
+class ChfIndexPresenter < Blacklight::IndexPresenter
+  class_attribute :description_max_chars
+  self.description_max_chars = 10
```
hackmastera: 10 seems too short; I think I heard @clu say '3 lines' at one point? Maybe an idea of a desired number of words would be more helpful; we could then use the avg English word length or something.

app/presenters/chf_index_presenter.rb
jrochkind: I am not truncating it in fedora, just upon display. I think that's the right thing to do.  This is the field that is used for display for description. If that were to change, then it would have to change. I don't know of any wy around that. 

If you had a _facet_ value that was hundreds of chars wrong, that would indeed be a problem, but not the one we're dealing with now.

app/presenters/chf_index_presenter.rb
jrochkind: oops, that was an accident for testing, I meant it to be 500. I can be whatever we like, and easily changed later. Will fix it to 500 for now.

app/presenters/chf_index_presenter.rb
hackmastera: Yeah that's not what I meant. I guess; where is this configuration? where do i go to look at index_fields? I want to take a look.

jrochkind: You and me both, buddy!  It's all over the place. But `configuration.index_fields` comes from the configuration in CatalogController

jrochkind: I was trying to avoid yet another place that has a `solr_name` call that has to match exactly the calls in the two other existing place, all of which have to be changed and kept in sync every time one changes. I am not happy with the hydra architecture here, but this is an attempt to avoid that.

hackmastera: ah, okay! I see it there. thanks.

I was concerned that in the model a property may be defined with more than one index field; how do we know which index field we're getting? But looking at the CatalogController I see that's not an issue.

PR #388, opened 2017-03-13
Place facet

### LINKED ISSUE #346: Place facet
Place facet should draw from 3 different properties; values should probably be copied into a 4th field to be used for faceting.

jrochkind:
Copies four place-related properties over into a solr field for facet

Closes #346


jrochkind:
will require a re-index.

PR #392, opened 2017-03-15
WIP, Q: Strip blank spaces before or after a heading fixes #385

### LINKED ISSUE #385: Strip blank spaces before or after a heading
@catlu had found that there were headings (subject and possibly others) which have been mismatched in Solr due to spaces at the start of end of a phrase ex: " Beckman" "Beckman" and "Beckman ". 

Would it be possible to have the form automatically strip leading or trailing spaces in a string? It doesn't seem like those would ever been meaningful enough to hold onto.

@hackmastera @jrochkind Do you have any thoughts on the difficulty of adding this?

jrochkind:
Q: Right now the migration rake task strips trailing/leading spaces from
ANY fedora property that is a String. Is this bad? Should I limit it to a certain
whitelist? If so, is there any place in existing code I could pull such a whitelist
of "properties that are editable in the form"?

### DIFF
--- a/lib/tasks/data_fixes.rake
+++ b/lib/tasks/data_fixes.rake
@@ -1,3 +1,4 @@
+require 'chf/data_fixes/util'
 require 'chf/data_fixes/work_dates'
 require 'chf/data_fixes/add_catlu_access'
 require 'chf/data_fixes/strip_strings'
@@ -6,6 +7,20 @@ namespace :chf do
   # Most of these are anticipated to be one-time fixes, may get removed
   # from rake later?
   namespace :data_fix do
+
+
+    desc "Bulk fix a variety of things"
+    task :bulk_fix => :environment do
+      # Apply a variety of fixes in one pass, to go quicker.
+      # what fixes are included may change from time to time,
+      # most of these are run-once things.
+      CHF::DataFixes::Util.update_works do |w|
+        [ CHF::DataFixes::WorkDates.new(w).change,
+          CHF::DataFixes::AddCatluAccess.new(w).change,
+          CHF::DataFixes::StripStrings.new(w).change
+        ].any?
+      end
+    end
 
     desc "Fix Work dates to be Dates"
     task :fix_work_dates => :environment do


hackmastera:
I can't think of a scenario in which that would be bad; let me look more thoroughly through some of the fields and get back to you.

app/forms/concerns/work_form_behavior.rb
```diff
@@ -68,6 +68,16 @@ def model_attributes(params)
       #  -- they don't have these fields in the form at all so they don't get repopulated.
       clean_params = encode_physical_container(params, clean_params)
       clean_params = encode_external_id(params, clean_params)
+
+      clean_params.keys.each do |key|
```
hackmastera: I wonder if we want to do this earlier; below the call to super?

Also it might be nice for this to be its own method; this one is getting a little long.

lib/chf/data_fixes/util.rb
```diff
@@ -0,0 +1,29 @@
...
+      # Generic utility code for updating all GenericWorks, with a progress
+      # bar.  Caller must yield a block that takes a single GeneircWork, possibly
+      # updates it in place (without #save'ing), and returns true only if it
+      # was updated.
+      def self.update_works
+        works_updated = []
+        progress_bar = ProgressBar.create(total: GenericWork.count, format: "|%B| %p%% %e %t")
+
+        GenericWork.find_each do |w|
+          updated = yield(w)
```
hackmastera: Happy to see this refactor and making it yield a block is a great idea.

jrochkind: Yep! It was always my plan, but I wanted to wait until I had 2-3 examples before extracting anything out, to make sure my abstraction met the actual design needs and wasn't premature over-engineering.

app/forms/concerns/work_form_behavior.rb
jrochkind: The example from awead did it after super, and your existing transformations were after super too! So after super seemed the way to go. 

I have no real opinion on whether extracting to a method in this case will make it more or less readable/maintainable. But I plan to add more stuff to the `clean_params.keys.each` loop (for description formatting), so I want to leave that loop in the main method. 

What do you think about waiting until we add a bit more stuff, and then refactoring? I also wasn't sure what the existing stuff did exactly, so wasn't sure how to refactor it.

hackmastera: Yes i meant after super but maybe before the other ones.

We can wait, sure.

PR #395, opened 2017-03-16
WIP description field formatting

### LINKED ISSUE #327: Description field: data cleanup and presentation
- [x] Replace instances of `<p>` with "\r\n\r\n"
- [x] Whitelist selected html tags at save time (render time maybe?)
- [x] Render field as html_safe (whitelisted html tags).
- [x] Convert line breaks at display time

Discussed with @jrochkind

jrochkind:
closes #327

- [x] HTML-aware truncation needed

Okay,this doesn't yet do any data migration, but is complete for new data. 

Before considering what data migration we may need, let's consider
what this will do to existing data. 

Right now, if you have html tags in your description, they'll be exposed as
HTML source:

~~~
<p>Let's say <em>I</em> wrote <ul><li>this</li></ul>.</p>
~~~

It'll show up just like that. 

After this change, the tags will be stripped out _on display_ to the public,
if someone looks at the record, it will look like this:

"Let's say I wrote this."

if they open it up in the editor, they will see the original code with tags, **BUT** if they save it without changing it, it will strip out the tags on save too, they'll be gone forever. Or they can edit it to be right and save it of course. 

If existing text has _allowed_ html tags, they'll stay there, won't be stripped out. 

If existing text has newlines, they'll be properly converted. So for instance if something had `<p>` tags _and_ double-newlines seperating paragraphs, it'll show up just right -- p tags will be stripped out, but newlines will be converted back to p tags. 

So this isn't actually that terrible. It depends on the nature of the data, and the nature of our stakeholder expectations for how much we have to do to accomodate their erroneous unsupported past use -- and how likely someone is to open up an edit page, and save it again without noticing
the description needs to be fixed first. 

From production:
* Number of records with `<p>` in description: 22
* Number of records with `<em>` in description: 2
* Number of records with `<strong>` in description: 0  
* Number of records with `<table>` in description: 0 (phew)

So we may not have to do any migration at all?


jrochkind:
Also I'll note I don't really know if the way i've implemented is the "best" way to do these things, but  the hydra stack is hard, this was the least-terrible one I found. But comments welcome.

config/locales/en.yml
```diff
@@ -42,7 +42,7 @@ en:
...
         inscription: "Use for text on a work that is captioned, dedicated, or inscribed. Input the inscription location (<strong><em>Ex: • Inside front cover • Verso bottom left • Back panel</strong></em>) and transcribe the text exactly as is on the work, including capturing misspellings, crossouts, or other anomalies as they appear."
-        subject: "Topical terms describing the “aboutness” of the work. Depicted persons, places, and time periods are covered here as well. Field supports FAST (derived from Library of Congress authorities) autocomplete. <strong><em>Ex: • Alchemy • Integrated circuits</strong></em>. Recommended."
+        subject: "Topical terms describing the “aboutness” of the work. Depicted persons, places, and time periods are covered here as well. Field supports FAST (derived from Library of Congress authorities) autocomplete. <strong><em>Ex: • Alchemy • Integrated circuits</strong></em>. Recommended. Allowed HHTML tags: b, i, cite, a."
```
hackmastera: Looks like you added this onto subject instead of description.

app/helpers/description_formatter_helper.rb
```diff
@@ -0,0 +1,32 @@
...
+
+    # a sufia helper to turn URLs into links.
+    text = iconify_auto_link(text)
+
+    text
+  end
+
+
+  def format_description_for_index(field)
+    if field.is_a? Hash
```
hackmastera: Interesting; when is it a hash? Does it come from solr that way?

spec/concerns/forms/work_form_behavior.rb
```diff
@@ -83,6 +83,17 @@
         expect(subject[:title]).to eq([stripped_title])
       end
     end
+
+    context "html in description field" do
+      let(:original_input) { '<p>Paragraph <b style="somestyle">bold</b> <a href="http://example.com">link</a><script>script</script></p>' }
```
hackmastera: Can you either modify this spec or add another spec with invalid html? I know we have some records where someone has been using `<p>` as if it were `<br />` (i.e. opening but never closing) and it'd be nice to have assurance that will be resolved.

hackmastera:
I think these numbers are small enough that if we can give @catlu the IDs (at least for the `p` instances; the `em` ones are actually searchable) she can fix them on a case-by-case basis. How did you get these counts; do you think you can repeat it to pull out ids for her?

jrochkind:
I found the counts just by doing `GenericWork.find_each { |w| w =~ /<em>/}` etc in a console.

app/helpers/description_formatter_helper.rb
jrochkind: No, it's not coming from Solr, it's some thing in Blacklight. I don't totally understand it either, just copied it from existing BL/sufia code.

hackmastera:
so to wrap up this plan: I think we should still do the first item on the issue, 
> Replace instances of <p> with "\r\n\r\n"

The 2 records with `<em>` I'm comfortable asking @catlu to change manually to `<i>`; those will be easy to find just searching for 'em'.

Otherwise this should take care of it.

jrochkind:
Okay, html-aware formatting, and a data fix task for `p`, `strong`, and `em`. 

I think maybe it's okay to run this on production without downtime?  It should only take 20 minutes since it only effects a few records, and unlikely to conflict with anyone.

catlu:
@hackmastera the two records I found when searching for 'em' were these two:
https://hydra.chemheritage.org/concern/generic_works/c821gj78w
https://hydra.chemheritage.org/concern/generic_works/c534fn96k

so I didn't edit. If you guys found different ones, just let me know.

jrochkind:
@catlu we haven't run our automated cleanup yet. You're welcome to change those manually, or else our automated cleanup should catch them, either way is fine.

catlu:
@jrochkind these two records don't have `em` as markup. They're being used in phrases like "got 'em" and "count 'em". Automated cleanup might mistakenly correct them? Unless @hackmastera wasn't actually talking about keyword search, in which case I've got the wrong records.

jrochkind:
@catlu  okay, cool, should not be a problem, the automated cleanup should not touch them.

jrochkind:
I believe there are two records with actual `<em>` tags. I didn't record the IDs. I can find them again if you like, or we can just let the automated cleanup fix them. The automated cleanup is written such that it should not disturb files that use the word or string 'em' not surrounded by brackets as an html tag.

catlu:
Ah, got it. Automated cleanup works for me.

hackmastera:
Looks good to me; @jrochkind feel free to merge whenever you're ready to run the update so we don't lose tags we would want to convert in the meantime.

jrochkind:
Updated 21 records, total scan took under 4 minutes hooray. 

~~~
hydep@lakitu-prod:/opt/sufia-project/current$ RAILS_ENV=production bundle exec rake chf:data_fix:description_html
|=================================================| 100% Time: 00:03:43 21 saved
Updated 21 works
~~~

PR #397, opened 2017-03-20
Update fields displayed for search results

### LINKED ISSUE #330: Update fields displayed for search results
Specifications needed.

jrochkind:
closes #330

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -114,8 +114,9 @@ class CatalogController < ApplicationController
     # solr fields to be displayed in the index (search results) view
     #   The ordering of the field names is the order of the display
     config.add_index_field solr_name("title", :stored_searchable), label: "Title", itemprop: 'name', if: false
+    config.add_index_field solr_name("additional_title", :stored_searchable), label: "Additional Title", itemprop: 'name', if: :present?
     config.add_index_field solr_name("description", :stored_searchable), label: "Description", itemprop: 'description', helper_method: :iconify_auto_link
-    config.add_index_field solr_name("subject", :stored_searchable), label: "Subject", itemprop: 'about', link_to_search: solr_name("subject", :facetable)
+    config.add_index_field solr_name('date_of_work', :stored_searchable), label: "Date", itemprop: 'date'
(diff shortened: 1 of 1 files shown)


app/controllers/catalog_controller.rb
```diff
@@ -114,8 +114,9 @@ def self.chf_search_fields
     # solr fields to be displayed in the index (search results) view
     #   The ordering of the field names is the order of the display
     config.add_index_field solr_name("title", :stored_searchable), label: "Title", itemprop: 'name', if: false
+    config.add_index_field solr_name("additional_title", :stored_searchable), label: "Additional Title", itemprop: 'name', if: :present?
```
hackmastera: what is itemprop? I think I thought it was the same as the property name as defined on the model, but clearly I was wrong.

jrochkind: I honestly don't really know what itemprop is. I'm not sure it's used for much of anything. I think it's supposed to be used for schema.org metadata markup generation (that's where I recognize the term 'itemprop' from), but does not seem to be doing this in our app. It's kind of another mystery of the hydra stack to me at the moment.

jrochkind: But you're right it was an accident to leave it as a duplicate of the one from 'title', that was a copy paste error. I guess I'll change it to "additional_title", I dunno.  I could spend time figuring out what hydra intends it to be doing and if it's doing that and if not why not, but I could spend many weeks doing that for all parts of hydra, ya know?

jrochkind: Actually maybe I'll just leave it as it is, `name` is the closest schema.org, and as far as I know is repeatable. Really don't know how this stuff works, will pay more attention to it when doing the hoped-for future overhaul of the entire search results page. https://project-hydra.slack.com/archives/C0F9JQJDQ/p1490103783658224

hackmastera: Ha, okay, so I have put all kinds of junk in itemprop inappropriately. I'll open an issue to revise those.

jrochkind: > Ha, okay, so I have put all kinds of junk in itemprop inappropriately. I'll open an issue to revise those.

Seriously, I don't think it matters, I don't know it's a good use of time to deal with that now. I'd really like to entirely redo this page, when we have a chance, and deal with it all then. But do you what you feel is best of course!

PR #398, opened 2017-03-20
Display additional titles with main title in browse view

### LINKED ISSUE #377: Display additional titles with main title in browse view

jrochkind:
Closes #377

screen shot:

screen shot 2017-03-20 at 6 30 03 pm


jrochkind:
I think I wanna redo this entire page, so didn't want to spend too much time on getting it just right now. But prob not til after the soft launch. 

I'm not sure; h4 looked more or less right but you're right it's not ideal in terms of html semantics. I think there are few actual screen readers or assistive technology that are _actually_ confused by this, it's common enough in the wild that assistive technology has had to work around it. 

But I can make it an h2 and add CSS styling to make it look right anyway if you'd prefer.

hackmastera:
> But I can make it an h2 and add CSS styling to make it look right anyway if you'd prefer.

I think that would be better!

jrochkind:
Okay, I'll just have to remember to _remove_ the styling again when this page is totally redesigned. 

I haven't quite figured out good practices for keeping my CSS well-organized and maintainable, although I keep trying. It's easy to end up with CSS that no longer applies to anything in the app but is still bloating your CSS file(s).

hackmastera:
Truth.

jrochkind:
ok, css restructured to make it easier for me to add the css in a consistent way, and additional title is now h2 styled like an h4 instead of an h4.

PR #399, opened 2017-03-20
Change Genre term "Records (Documents)" to "Documents"

### LINKED ISSUE #201: Change Genre term "Records (Documents)" to "Documents"
Will need deletion/addition of term to associated Hydra records.

jrochkind:
Fixes #201

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -130,7 +130,7 @@ module Chufia
       'Press releases',
       'Prints',
       'Rare books',
-      'Records (Documents)',
+      'Documents',
       'Sample books',
       'Scientific apparatus and instruments',
       'Slides',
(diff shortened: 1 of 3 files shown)


jrochkind:
I guess the other option would be leaving the old bad one in the list for now, merging now, and making another commit to remove the old bad one from the list only after we've run migration. 

I am not super happy with either alternative; I'd like to find a way we don't have to remember to do something later, but haven't come up with that yet. 

What do you think?

hackmastera:
Either is really fine with me except that I still prefer to run with downtime until we're on sufia 7.3 and have some record locking. So maybe it's simpler to just do it all at once, waiting to merge until we're ready.

PR #400, opened 2017-03-21
Change UI label "Maker" to "Creator"

### LINKED ISSUE #363: Change 'maker' to 'creator'
basically a full-on find / replace

jrochkind:
Closes #363

@hackmastera, I think these are the only places to change the UI label "Maker"
to "Creator", my understanding of #363 is we just wanted to change ui labels, not
any underlying field names. 

Is this what you expect for #363?

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -97,7 +97,7 @@ class CatalogController < ApplicationController
     # solr fields that will be treated as facets by the blacklight application
     #   The ordering of the field names is the order of the display
     config.add_facet_field solr_name("subject", :facetable), label: "Subject", limit: 5
-    config.add_facet_field solr_name("maker_facet", :facetable), label: "Maker/Creator", limit: 5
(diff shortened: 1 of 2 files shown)


hackmastera:
It's really not in there all that much is it.

I'm having a sudden thought that we do have a field `creator_of_work` which is labeled 'creator'. @catlu I assume you thought about that already? Maybe this isn't an issue since I believe that field was also intended as a bit of a catch-all for anything we didn't have a more specific term for; semantically in line with the heading.

In terms of end-users it's really just the facet that we're labeling as 'Creator' instead of 'Maker.'

For our cataloging users the form field is also being relabeled. Again, this means that there will be an option 'creator' in the dropdown under the field heading 'Creator'. It will just require a bit of extra clarity when we communicate about the field vs. the field grouping.

@catlu I defer to your approval!

catlu:
Yeah, no issues there! The creator field is meant as a catch-all the way @hackmastera described, so having the field label reflect that makes sense.

PR #402, opened 2017-03-21
Super fast reindex_works tasks

jrochkind:
* Only does works to prevent the slower find all fedora ids task
* batches solr adds
* uses a progress bar

### DIFF
--- a/lib/tasks/chf_tasks.rake
+++ b/lib/tasks/chf_tasks.rake
@@ -56,6 +56,22 @@ namespace :chf do
     puts "total: #{reporter.matches.size}"
   end
 
+  desc 'Reindex all GenericWorks'
+  task reindex_works: :environment do
+    # Like :reindex, but only GenericWorks, makes it faster,
+    # and allows us to add a progress bar easily.
+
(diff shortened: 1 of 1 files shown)


lib/tasks/chf_tasks.rake
```diff
@@ -56,6 +56,31 @@ namespace :chf do
...
+    add_batch_size = ENV['ADD_BATCH_SIZE'] || 50
+
+    progress_bar = ProgressBar.create(:total => GenericWork.count, format: "%t: |%B| %p%% %e")
+    solr_service_conn = ActiveFedora::SolrService.instance.conn
+    batch = []
+
+    GenericWork.find_each do |work|
+      batch << work.to_solr
+
+      if batch.count % add_batch_size == 0
```
hackmastera: What happens to the last set of records if the total isn't a multiple of 50? Do we need an extra add outside of the loop to catch those?

jrochkind: oops good catch!

jrochkind: i love having a peer reviewer! didnt have that at jh really

hackmastera: Yeah having people review your code is great!

PR #406, opened 2017-03-22
cap tasks for running tasks, including with maintenance mode

jrochkind:
TASK=chf:data_fix:library_division REASON="testing things" UNTIL="12pm Eastern Time" bundle exec cap staging invoke:rake:with_maintenance

Will turn maint mode on for you, run task, turn it back off -- even if
task fails. If you want to leave it on if task fails, SAFE_MAINT=false. 

Or to just run a rake task on a remote server without maint mode:

    cap staging invoke:rake TASK=chf:data_fix:whatever

### DIFF
--- a/lib/capistrano/tasks/invoke.rake
+++ b/lib/capistrano/tasks/invoke.rake
@@ -1,0 +1,104 @@
+namespace :invoke do
+
+  #     cap <stage> invoke:rake TASK=chf:data_fix:something[,other:task]
+  desc "Execute a rake task on a remote server"
+  task :rake do
+    if ENV['TASK']
+      tasks = ENV['TASK'].split(',')
+
+      on roles(:app) do
+        within current_path do
+          with rails_env: fetch(:rails_env) do
+            tasks.each do |task|
+              # warning, may be executing this on multiple servers if we have
+              # multiple 'app' servers later, which would be bad.
+              # Will have to deal with that then, not sure best way.
+              execute :rake, task, interaction_handler: CHF::CapistranoHelp::StreamOutputInteractionHandler.new(:stderr)
+              info("finished rake #{task}")
+            end
+          end
+        end
+      end
+
+    else
+      # Not really sure why we can't just use `error` method, maybe not since
+      # cap 4?
+      SSHKit.config.output.error "\n\nFailed! You need to specify the 'TASK' parameter!\n" +
+           "Usage: cap <stage> invoke:rake TASK=your:task[,other:task]"
+    end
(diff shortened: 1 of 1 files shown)


jrochkind:
@hackmastera I'm going to merge this to master so I can try it tonight, but please feel free to review anyway, it can be reversed out, or modified. 

It only adds some new cap tasks, so i judge it safe to merge into master so I can use it on production tonight. Hope that's okay!

lib/capistrano/tasks/invoke.rake
```diff
@@ -0,0 +1,104 @@
...
+# else and `require` it, not sure why. That'd be better.
+
+# based on sshkit's MappingInteractionHandler, but all
+# we want to do is log it as we get it! Not really an interactin handler at all,
+# just a stream logger.
+module CHF
+  module CapistranoHelp
+    class StreamOutputInteractionHandler
+
+      # set log level to :stderr, and it will be written directly to stderr console
```
hackmastera: This seems like overloading log_level with a destination rather than an actual log level? Which I don't really have a problem with; if it becomes an issue we could always split it out. I'm just asking since I don't think it's a convention I've seen before.

lib/capistrano/tasks/invoke.rake
```diff
@@ -0,0 +1,104 @@
...
+      # instead of capistrano logging, which works to get byte-by-byte output
```
hackmastera: I'm a little confused by the grammar of the 'which' here -- writing directly to stderr allows the byte-by-byte output is what I think you're saying but it reads as if capistrano logging allows the byte-by-byte output. So I think a small edit for clarity in this comment would help.

lib/capistrano/tasks/invoke.rake
jrochkind: Yes. I think for as simple a class as this one, it makes more sense to just do that (and doc it in comments as I tried to do), rather than add complexity to the class. I think it's a reasonable API `StreamOutputInteractionHandler.new(:stderr)`, or `.new(:info)`. It could be argued that logging levels are a kind of destination too (some loggers send differnet levels to differnet places!). 

Mostly though I don't expect this to be used anywhere outside this file really and I don't think it matters much.

lib/capistrano/tasks/invoke.rake
jrochkind: You understand exactly. If you tell me what language makes more sense to you happy to paste it in, or you're welcome to commit it yourself of course!

PR #407, opened 2017-03-27
Upgrade sufia from 7.2 to 7.3 and ActiveFedora to 11.1.5

hackmastera:
For review / comment.

I'd like to deploy to staging and ask some users to test before deploying to production.


app/controllers/catalog_controller.rb
```diff
@@ -2,7 +2,6 @@ class CatalogController < ApplicationController
   include Hydra::Catalog
   include Hydra::Controller::ControllerBehavior
   include Sufia::Catalog
-  include BlacklightAdvancedSearch::Controller
```
jrochkind: are you intentionally removing AdvancedSearch-like behavior here?  Is it related to the Sufia upgrade?

app/forms/batch_edit_form.rb
```diff
@@ -57,6 +57,9 @@ def self.model_attributes(params)
     params[:rights] = Array(params[:rights]) if params[:rights].present?
     clean_params = super #hydra-editor/app/forms/hydra_editor/form.rb:54
     clean_params = encode_external_id(params, clean_params)
+    # These are removed due to a bug released in sufia 7.3:
+    # https://github.com/projecthydra-labs/hyrax/issues/652
+    clean_params['permissions_attributes'] = params['permissions_attributes'] if params['permissions_attributes']
     clean_params
```
jrochkind: I'm worried the clean_params are supposed to be making sure someone without the correct privs can't set this, and this workaround might mean anyone can set this if they hand-craft an HTTP request. Which would be especially dangerous for setting permissions!

Is my concern possibly valid?

app/jobs/batch_create_job.rb
```diff
@@ -1,51 +0,0 @@
...
-      end
-    end
-
-    # Override this method if you have a different rubric for choosing the model
-    # @param [Hash] attributes
-    # @return String the model to create
-    def model_to_create(attributes)
-      Sufia.config.model_to_create.call(attributes)
-    end
-end
```
jrochkind: background on why this file is being removed? Sorry, I tried to find the commit in this PR that actually removed it, but apparently my git knowledge isn't enough to figure out how to do that!

config/deploy.rb
```diff
@@ -51,4 +51,20 @@
...
   before :restart, :resquepoolrestart
 
+  # load the workflow configs
+  desc "Load workflow configurations"
+  task :loadworkflows do
+    on roles(:web) do
+      execute :rake, 'curation_concerns:workflow:load'
+  end
+  before :restart, :loadworkflows
+
```
jrochkind: what the heck is this 'workflow configuration' stuff, new in new version of Sufia/CC? Doc'd anywhere you know of?

lib/chf/import/credit_builder.rb
```diff
@@ -1,33 +0,0 @@
-module Chf
```
jrochkind: what's with this going away? Related to update, or just something we weren't using anymore anyway? Same for DateOfWorkBuilder.

app/controllers/catalog_controller.rb
hackmastera: Sufia doesn't expose any advanced search in its ui as far as I know. the choices were to remove the advanced search stuff from this controller or add the advanced search gem into our Gemfile. It's good to call this out as something to double-check in testing.

app/forms/batch_edit_form.rb
hackmastera: Let me look at this more; there may be another way I can get around the upstream bug.

app/jobs/batch_create_job.rb
hackmastera: This was an override for a fix that was resolved upstream; https://github.com/chemheritage/chf-sufia/commit/b541309041aab600e5a40296d407b0468f720f1a

lib/chf/import/credit_builder.rb
hackmastera: This was related to the upgrade from sufia 6.x to sufia 7. The code it built on only exists in the sufia 7.2-migration branch, so it is both no longer needed and no longer working (the tests fail).

config/deploy.rb
hackmastera: Admin Sets and Workflows are new functionality in sufia 7.3. Personally I don't think it's quite mature enough to start using. But you do have to have defaults in place for the code to work. Workflow documentation is forthcoming; Jeremy just said he has it slated for next week (many at ldcx also asking for it).

app/controllers/catalog_controller.rb
jrochkind: sounds good, we can always add it back in if we want/need it. I'm down with keeping things as simple as we can get away with for soft launch.

lib/chf/import/credit_builder.rb
jrochkind: Hooray for removing code we're not going to use anymore!

app/forms/batch_edit_form.rb
hackmastera: Here is a [debug gist](https://gist.github.com/hackmastera/f169c5b2e36373bbd0dd226fe0d9b1c0). I don't see any permissions checking, really, except the buggy implementation at `sufia-7.3.0/app/forms/sufia/forms/work_form.rb @ line 50` which is what is breaking this for batch edit (a batch edit form doesn't have an admin set).

I think permissions checking stuff usually happens during persistence; I'm seeing evidence of it in the actor stack. Makes me a bit confused why the bug was introduced this way to begin with. I'll ask about it on the hyrax ticket (https://github.com/projecthydra-labs/hyrax/issues/652)

I can think of a couple other ways to get around this, but they more or less do the same thing, which is circumnavigate the [upstream sanitize_params](https://gist.github.com/hackmastera/f169c5b2e36373bbd0dd226fe0d9b1c0#file-gistfile1-txt-L39) which is causing the trouble, and none of them are prettier.

jrochkind: The word "sanitize" still makes me suspect this was meant to apply permissions controls; but I could be wrong and it never was, or the intent/design could be confused by multiple developers intending different things with it. 

But if this seems the best way to do it, that seems fine!  I _think_ this entire controller action should be protected at least to logged-in users, and all logged in users at present are staff, who can be trusted not to try to hack the system, so it's definitely fine at least for now. I'm good.  Working with sufia definitely involves lots of "good enough for now".

hackmastera: Okay, it's true; let's stick with this then as good enough for now.

hackmastera:
After deploy, need to run the reindex collection job. Other jobs needed for this upgrade have been written in to capistrano.

jrochkind:
Neat. Can you give me a link to the jobs written in to cap, so I can see how to do that? Any way to put the reindex collection one in there too?

jrochkind:
Are there any other dependencies we might want to update, as long as we're updating and testing?

Like blacklight, we're currently on 6.7.2, 6.8.0 is most recent, doens't look like it _should_ have any backwards-breaking changes. https://github.com/projectblacklight/blacklight/releases

Any other gems we might want to too?

Gemfile
```diff
@@ -3,6 +3,9 @@ source 'https://rubygems.org'
 # sufia stuff!
 gem 'sufia', '7.3.0'
 gem 'blacklight_range_limit', ">= 6.1.2"
+# It may seem like this is a dependency upstream, but we need it here.
+# https://project-hydra.slackarchive.io/dev/page-100/ts-1491587437169731
+gem 'rsolr', '~> 1.0'
```
jrochkind: this slack link sadly won't work, because hydra slack only keeps limited history since it's a free slack account. It would be nice to able to link to slack conversations in comments! There's some slack archive service whose URLs you can use as a workaround, but too much trouble for me too.

PR #413, opened 2017-03-27
'Medium' facet should only show for logged-in users

### LINKED ISSUE #364: 'Medium' facet should only show for logged-in users
but keep it as a facet so we can link to any given set

jrochkind:
Closes #364


hackmastera:
nice!

jrochkind:
Ha, yeah, it took me like 2 hours to figure out those like 20 characters! And I'm still not sure why it works or if it's the "right" way to do it, but it works, good enough for now!

PR #414, opened 2017-03-29
Remove file hint

### LINKED ISSUE #286: if you upload a new version of a file the help text still says "A PDF is preferred."

jrochkind:
Formerly "A PDF is preferred"

Closes #286

I was not able to actually find where the upload a new version is in
the UX, although I know I found it last week! So have not confirmed, but
pretty sure this will remove the prompt.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -28,6 +28,8 @@ en:
         date_of_work: "Date"
         description: "Description"
     hints:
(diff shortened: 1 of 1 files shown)


hackmastera:
You have to go to a file (as opposed to a work), click edit, and then go to the 'versions' tab. Anyway, this works; merging!

PR #415, opened 2017-03-29
Solve "Single and Batch Work upload forms too similar in Sufia 7" closes #261

### LINKED ISSUE #261: Single and Batch Work upload forms too similar in Sufia 7
From Slack: "that said, the difference between the two forms is really similar, the only way to tell is the tabs are switched so at some point renaming the heading of the forms (both say "add new work" currently) might be helpful"

Maybe renaming the headline would be a simple solution?

jrochkind:
1) Different title for single vs batch work upload form. Not sure if there's a better way to organize the i18n keys, this seemed clever to me, or might be too confusing.

2) put batch upload files instructions in alert style to further differentiate batch from single work upload

### DIFF
--- a/app/assets/stylesheets/local/upload_works.scss
+++ b/app/assets/stylesheets/local/upload_works.scss
@@ -1,0 +1,3 @@
+.new_batch_upload_item .instructions {
+  @extend .alert, .alert-info;
+}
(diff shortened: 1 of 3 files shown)


hackmastera:
looks good to me; took a screenshot so @catlu can approve. If she likes I'm +1 to merge.

screen shot 2017-04-03 at 1 20 56 pm

catlu:
I like!

PR #422, opened 2017-04-03
Add configuration for working_path

hackmastera:
This location used as a holding area for files pulled from fedora for
derivatives generation. Generally not needed during ingest, since the
files are taken from their upload location (managed by CarrierWave).

Sufia made an effort at one point to make sure all CC config values were
managed by Sufia's config, as well, but this one must have slipped
through the cracks. Hence it's in a new CC initializer.

### DIFF
--- a/config/initializers/curation_concerns.rb
+++ b/config/initializers/curation_concerns.rb
@@ -1,0 +1,6 @@
+CurationConcerns.configure do |config|
+  # Location on local file system where uploaded files will be staged
+  # prior to being ingested into the repository or having derivatives generated.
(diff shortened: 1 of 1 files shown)


jrochkind:
Cool. medium-longterm we could create our _own_ variables (global if needed) and then reference them in these various initializers, if we want to ensure all the various temp directories 'match' or are in the same parent directory and DRY the config. But not neccesary to merge this, I'm merging!

hackmastera:
Let's revisit that idea once we're on hyrax, since things will be somewhat simpler then.

Issue #329, opened 2017-02-17
Timeline search feature


hackmastera:
a date slider or some such

jrochkind:
https://github.com/projectblacklight/blacklight_range_limit is hopefully working with the version of Blacklight and everything else we got.

hackmastera:
I'd like to talk about this one together.

catlu:
In the "my works" view, the date slider isn't there, instead it's a listing of dates:

![capture 1](https://cloud.githubusercontent.com/assets/11876037/24866330/cf963e28-1dd7-11e7-85bc-994267f8cb8c.PNG)

jrochkind:
It's astoundingly confusing to try and get this to work on My Works view. 

I don't like leaving it there, there could be thousands of unique dates, it can't be left as it is. 

If I can't get it to work in My Works view, what do you think about removing this facet from My Works view?  (Have to figure out how to do _that_ too, it's not obvious)

jrochkind:
Some notes on this feature in general and how it calculates things for 'odd' dates should be consulted at #440

catlu:
@jrochkind Either removing it or collapsing it is fine with me--I'm really not too fussed about it not working correctly in the My Works tab.

hackmastera:
@jrochkind ready to close?

jrochkind:
a timeline search feature is merged to master; there is no date facet/timeline search on 'my works' page, just public page.

jrochkind:
Weird. 

> clu [10:59 AM] 
> btw @jrochkind i noticed the date slider facet on production when i'm logged in, but it's not there when i'm logged out, is that intentional?

catlu:
another weird thing: when I facet to "unknown" in the date slider for production data, lots of stuff with specific dates show up.

jrochkind:
Weirder, I can't reproduce this in dev on my workstation, even though I saw it happening in production. 

Wait, now I can't reproduce in production either. @catlu can you still reproduce in production? Try refreshing the page?

jrochkind:
Ah, I think production needs a reindex now that the date facets are live, to get date facet values in there. Will run now, it does not require downtime and should not interfere with concurrent staff use.

catlu:
@jrochkind Can confirm things look good to me. We do have a good amount of stuff still waiting in the wings for metadata, so that's definitely related to workflow and non-public visibility, so not a bug!

PR #440, opened 2017-04-06
WIP date range facet/slider

jrochkind:
This needs to be evaluated by stakeholders for UX -- either we need to reserve staging for a few days for that, or we can just deploy to production and get eval there before closing issue #329

Some oddities to keep aware of:

* The potential date input is various, and includes possible ambiguous or erroneous dates, or date information it's hard to represent completely accurately in a timeline/facet limit. 
  * If a date is marked 'before', 'after', or 'circa', it's indexed for the limit the same as if it had no qualifier at all. "before 1950" is put on the timeline at 1950. 
  * If a date has a 'start' of "decade" or "century", the "finish" is _ignored_.  Something entered as "Start: 1950s, End: 1962", will jut be indexed as "1950s", ie 1950-1959.  
  * A finish _before_ a start is ignored, just the start is indexed. 
  * Something with a qualifier but no date is ignored. Something with "Undated" qualifier but also a date is treated as just "Undated" and not put on the timeline. 
  * Date entry that isn't in `YYYY(-mm(-dd))` format is just ignored. 

* Date ranges are a bit weird for the histogram summary. For instance. Recall that the histogram summary appears for the _current result set_ to let you limit further, if you've already limited with a query or facet choices, the date facet histogram summary characterizes the current result set. i'll choose some 'extreme' examples to illustrate the issue.
  *  Imagine your current result set includes 100 documents, all of which are indexed as 1900-1950. The histogram will make it look like there are 100 documents in each date segment, but when you click on them, it's the same 100 documents each time.  
  * Or, imagine the result set includes those same 100 documents indexed 1900-1950, but _another_ 100 indexed as 1951.  The histogram will make it look there are an equal number of documents throughout the timeline, but in fact if you limit to just 1951 you get 100 distinct documents, but limit to any other year in the range you get the other 100 documents. 

Additionally there's a small UI bug in the blacklight_date_range gem that I want to try to fix.

### DIFF
--- a/app/models/search_builder.rb
+++ b/app/models/search_builder.rb
@@ -1,9 +1,15 @@
 # frozen_string_literal: true
+
+# Note: In Sufia 7.x at all, this is _not_ used by the CatalogController,
+# which uses it's own `Sufia::CatalogSearchBuilder` instead. Unclear how
+# to customize it. This may be true of other sufia/hyrax searchbuilders as well.
+# https://github.com/projecthydra-labs/hyrax/issues/707
+#
+# See also .config/initializers/sufia_catalog_search_builder_overrides.rb
 class SearchBuilder < Blacklight::SearchBuilder
   include Blacklight::Solr::SearchBuilderBehavior
   include BlacklightRangeLimit::RangeLimitBuilder
 
   include Hydra::AccessControlsEnforcement
   include CurationConcerns::SearchFilters
-
 end
--- a/app/models/search_builder.rb
+++ b/app/models/search_builder.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 class SearchBuilder < Blacklight::SearchBuilder
   include Blacklight::Solr::SearchBuilderBehavior
+  include BlacklightRangeLimit::RangeLimitBuilder
+
   include Hydra::AccessControlsEnforcement
   include CurationConcerns::SearchFilters
 
--- a/config/initializers/sufia_catalog_search_builder_overrides.rb
+++ b/config/initializers/sufia_catalog_search_builder_overrides.rb
@@ -1,0 +1,16 @@
+# Unclear how to customize the Sufia::CatalogSearchBuilder, which
+# we need to do for blacklight_range_limit.
+# https://github.com/projecthydra-labs/hyrax/issues/707
+
+# to_prepare used in case there's dev-mode class reloading of the
+# thing we're patching. But we try to check to not do it multiple
+# times in case there isn't.
+Rails.application.config.to_prepare do
+  # would like to look up the class dynamically in case it changes,
+  # Sufia does this weird making this the only way, Hyrax is somewhat
+  # less weird.
+  klass = CatalogController.new.search_builder_class
+  unless klass.ancestors.include? BlacklightRangeLimit::RangeLimitBuilder
+    klass.send(:include, BlacklightRangeLimit::RangeLimitBuilder)
+  end
+end
--- a/spec/factories/generic_works.rb
+++ b/spec/factories/generic_works.rb
@@ -2,6 +2,7 @@ FactoryGirl.define do
   factory :generic_work, aliases: [:work, :public_work], class: GenericWork do
     transient do
       user { FactoryGirl.create(:user) }
+      dates_of_work { [DateOfWork.new(start: (1900 + rand(100)).to_s)] }
     end
 
     title ['Test title']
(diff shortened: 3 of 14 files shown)


jrochkind:
UI bug on tooltips fixed, waiting for merge at https://github.com/projectblacklight/blacklight_range_limit/pull/61

app/models/search_builder.rb
```diff
@@ -1,7 +1,15 @@
 # frozen_string_literal: true
+
+# Note: In Sufia 7.x at all, this is _not_ used by the CatalogController,
```
hackmastera: Still not sure what the implications of this are. Is there a place where the date facet will not be available?

hackmastera: oh, I see, you did the monkey patch to account for it.

config/initializers/sufia_catalog_search_builder_overrides.rb
```diff
@@ -0,0 +1,16 @@
+# Unclear how to customize the Sufia::CatalogSearchBuilder, which
+# we need to do for blacklight_range_limit.
+# https://github.com/projecthydra-labs/hyrax/issues/707
+
+# to_prepare used in case there's dev-mode class reloading of the
+# thing we're patching. But we try to check to not do it multiple
+# times in case there isn't.
+Rails.application.config.to_prepare do
+  # would like to look up the class dynamically in case it changes,
+  # Sufia does this weird making this the only way, Hyrax is somewhat
```
hackmastera: Should we log or raise something when we're on hyrax to make sure we change this?

spec/factories/generic_works.rb
```diff
@@ -2,13 +2,18 @@
...
+      dates_of_work { [DateOfWork.new(start: (1900 + rand(100)).to_s)] }
     end
 
     title ['Test title']
     visibility Hydra::AccessControls::AccessRight::VISIBILITY_TEXT_VALUE_PUBLIC
 
     after(:build) do |work, evaluator|
       work.apply_depositor_metadata(evaluator.user.user_key)
+
+      # unfortunately this does force saving the dates_of_work
```
hackmastera: Would it be better to create a date factory and use that here?

config/initializers/sufia_catalog_search_builder_overrides.rb
jrochkind: It'l keep working fine on Hyrax as far as I know, just could be better. I don't think it's needed personally, but judgement call.

spec/factories/generic_works.rb
jrochkind: I don't think so? But maybe? Can always be done later if we need it? I don't anticipate needing it.

hackmastera: It seems like this would allow use of build instead of create, and alleviate the slowdown. use it as a transient attribute? this is not a thing i have great knowledge of. i do think there are a lot of places where tests could be faster, though, so not hung up on it.

spec/indexers/generic_work_indexer/date_values_spec.rb
```diff
@@ -0,0 +1,86 @@
+require 'rails_helper'
+
+RSpec.describe GenericWorkIndexer::DateValues do
+  # Tried using 'build' instead of 'create', sadly didn't work, dates
+  # get thrown out. This does slow it down a lot cause Fedora. :(
```
hackmastera: maybe this comment is out of date, actually?

spec/factories/generic_works.rb
jrochkind: I don't think it would. All build does is give you an unsaved model, nothing magic about the models it gives you, you can create them youself too, FactoryGirl is just a convenience. 

What I discovered is if you give AF an un-saved date object and add it to the date_of_work relationship -- AF will save it at that point. Even though the 'parent' object is still not yet saved. 

FactoryGirl won't get around this. But it's not the end of the world, just another day in the life of AF. (Rails AR may do the same with associations in some cases).

spec/indexers/generic_work_indexer/date_values_spec.rb
jrochkind: I guess it is, I got it working with 'build'.  dates still get persisted though. Should prob elmiminate that comment, yeah.

jrochkind:
@hackmastera think this is ready for merge?

hackmastera:
@jrochkind can you change that one comment that is out of date?

spec/indexers/generic_work_indexer/date_values_spec.rb
jrochkind: removed comment

PR #446, opened 2017-04-10
Staff-only fields on show, closes #367

### LINKED ISSUE #367: Public/private fields
These fields don't need to be available to public:

- [x] Identifier
- [x] File creator
- [x] Admin note

jrochkind:
Note additional of global staff_user? helper method, recommend using this
instead of checking current_user.present? to express semantics better, and
allow easier changes in future.


jrochkind:
There oughta be a better way to mock helper methods for a view spec, but I couldn't make it work, so did it a hacky hacky way. shrug.

hackmastera:
Looks great! Merge conflicts, probably from your other pr.

app/views/curation_concerns/base/_attribute_rows.html.erb
```diff
@@ -1,3 +1,17 @@
+<<<<<<< HEAD
```
hackmastera: mergey merge

PR #456, opened 2017-04-11
Restrict certain search fields to be admin-only

### LINKED ISSUE #425: Admin notes should not be searchable to guest users
Only logged in users should be able to search this field.

I'm not sure how to accomplish this; this is initially an issue to discuss feasibility.

jrochkind:
Adding it to both places we add SearchBuilder configuration, cause
who knows what sufia/hyrax uses where, and it may change.

Closes #425

### DIFF
--- a/app/models/search_builder.rb
+++ b/app/models/search_builder.rb
@@ -12,4 +12,5 @@ class SearchBuilder < Blacklight::SearchBuilder
 
   include Hydra::AccessControlsEnforcement
   include CurationConcerns::SearchFilters
+  include SearchBuilder::RestrictAdminSearchFields
 end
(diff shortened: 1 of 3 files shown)


jrochkind:
Tests are probably a good idea, but I'm not sure what to test.  Test of the SearchBuilder itself risks false negative, if the architecture is refactored such that our tests still pass but it's not actually being used by the app to do what we want anymore. 

I guess an actual user-view browser test?  Those are a pain. But maybe is what we need to do?

jrochkind:
spent some time trying to make a SearchBuilder-based spec. I wasn't sure how reliable it would be anyway, I was having to assume so much about how the framework behaved -- and then I couldn't get it to work anyway, the architecture is too messy.

hackmastera:
okay I made an issue for integration tests https://github.com/chemheritage/chf-sufia/issues/466

PR #475, opened 2017-04-14
Cap fix

hackmastera:
It looked like capistrano wasn't correctly restarting our app on deploy. Update to a newer deploy strategy; upgrade the gem while we're at it. Although the gem hasn't been touched in a while; is it possible it's no longer maintaned? Has everyone switched to puma?

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -55,7 +55,7 @@ end
 group :development do
   gem 'capistrano', '3.4.0'
   gem 'capistrano-bundler', '1.1.4'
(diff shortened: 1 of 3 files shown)


sanfordd:
Is there a risk to being stuck with an older passenger version if there is nobody else on it, or is that just another in the list of Sufia vs Hyrax issues we'll need to keep in mind as people move to Hyrax?

hackmastera:
I'm not concerned about passenger; the gem I'm referencing here is a capistrano plugin that allows you to restart passenger.

PR #481, opened 2017-04-17
Fix batch permissions edit

### LINKED ISSUE #463: Batch editing: works were put under a lease when they should've been public visibility
Was batch adding user permissions to a group of works, the users were added but then the works were put under a lease when they should've been public visibility.

hackmastera:
Fixes #463 

The bulk of this is a revert of a pre-7.3 fix that was half-fixed upstream; the upgrade's fix conflicted with the previous fix. The rest of the PR adds back in desired enhancements.


app/controllers/batch_edits_controller.rb
```diff
@@ -7,4 +7,30 @@ def form_class
     BatchEditForm
   end
 
+  # nearly straight-up copy of sufia's method
+  def update
```
jrochkind: I assume "nearly" straight-up copy of sufia's method means it's not an _exact_ copy... or there'd be no reason to copy it and override it? Would it be good for the comment to say what the "nearly" is?

app/controllers/batch_edits_controller.rb
```diff
@@ -7,4 +7,30 @@ def form_class
...
+  # object needs to be saved before it's passed to the jobs.
+  def update_works
+    batch.each do |doc_id|
+      obj = ActiveFedora::Base.find(doc_id, :cast=>true)
+      previous_visibility = obj.visibility
+      update_document(obj)
+      obj.save
+      VisibilityCopyJob.perform_later(obj) unless previous_visibility == obj.visibility
+      InheritPermissionsJob.perform_later(obj) if work_params.fetch(:permissions_attributes, nil)
+    end
```
jrochkind: What is this doing differnetly than upstream?  Doing it in a bg job instead of foreground? Or propogating visibility and permissions at all? Other?

hackmastera:
@jrochkind I expanded the comments; let me know if this addresses your questions.

I suppose I could have done this as a monkey-patch; feel free to make an argument for that if you think it's better.

jrochkind:
I'm not really sure if a monkey-patch would be better, so am not going to make the argument, it's good enough! Nice work, thanks for figuring this out.

hackmastera:
:)

PR #484, opened 2017-04-18
Add maker field 'after', fixes #393

### LINKED ISSUE #393: New Maker dropdown label: After
For fine art cataloging where work is a copy of a known artist by an unknown one. The term is from Getty and ARLIS cataloging standards, but there's no predicate that corresponds, so mint our own?

hackmastera:


hackmastera:
It would be nice to switch and use Rails.configuration.makers in some of these places.

jrochkind:
I'm not familiar with `Rails.configuration.makers`, that's something we added locally, or a part of sufia, or other?

app/assets/javascripts/edit_metadata.js
```diff
@@ -20,6 +21,7 @@ chf.autocompletes = chf.autocompletes || {
   generic_work_place_of_publication: '/authorities/search/assign_fast/all',
   generic_work_place_of_creation: '/authorities/search/assign_fast/all',
   batch_edit_item_subject: '/authorities/search/assign_fast/all',
+  batch_edit_item_after:          '/authorities/search/assign_fast/all',
   batch_edit_item_artist:          '/authorities/search/assign_fast/all',
   batch_edit_item_author:          '/authorities/search/assign_fast/all',
   generic_work_addressee:          '/authorities/search/assign_fast/all',
```
jrochkind: Huh, i would not have known about this file, good to know! 

By autocompleting from 'assign_fast/all', does that mean someone _could_ autocomplete a subject as a creator/maker?

hackmastera: yes. There's not currently a way to query both personal and corporate names without querying the whole FAST vocabulary.

hackmastera:
`Rails.configuration.makers` would just pull from here https://github.com/chemheritage/chf-sufia/blob/master/config/application.rb#L41

Issue #439, opened 2017-04-05
Homepage category: Portraits & People


catlu:
Aggregate using metadata from:

- Genre
o   Portraits: 135

- Subject 
o   Portraits, Group: 16
o   Women in science: 31
o   Employees: 214

Issue #438, opened 2017-04-05
Homepage category: Science on Stamps


catlu:
Aggregate using metadata from:

-  Subject
o   Science on postage stamps: 119

Issue #437, opened 2017-04-05
Homepage category: Instruments & Innovation


catlu:
Aggregate using metadata from:

-  Genre
o   Scientific apparatus and instruments: 29

-  Subject
o   Artillery: 77
o   Machinery: 313
o   Chemical apparatus: 40
o   Laboratories--Equipment and supplies: 41
o   Chemical laboratories--Equipment and supplies: 4
o   Glassware: 22

jrochkind:
Does this mean it has to be in that _genre_ _and_ simultaneously be in one of those subjects? Or either is good enough, if it has one of those subjects but no genres at all or other genres, it's still included?

catlu:
@jrochkind should be that genre OR those subjects. The goal here was to include works at are instruments but filter out the large amount of Beckman materials that have that same subject heading.

catlu:
@jrochkind, I'm looking through this collection right now, but I don't see the museum instruments in here? Currently there should be 37 of them that fall under the scientific instruments genre.

jrochkind:
@catlu can you give me an example document that you believe should be showing up in the "Instruments & Innovation" category, but is not?

catlu:
Yep, the stuff from this search: https://hydra.chemheritage.org/catalog?_=1493746057393&f%5Bgenre_string_sim%5D%5B%5D=Scientific+apparatus+and+instruments&q=&search_field=all_fields

Issue #436, opened 2017-04-05
Homepage category: Alchemy


catlu:
Aggregate using metadata from:

- Subject
o   Alchemy: 19

jrochkind:
@catlu There's at least one record in production with subject `Alchemists` but not `Alchemy`. Add that subject too? 

But these will be easily changeable at any time, I think.

catlu:
Good catch @jrochkind. Yes, please add that one!

PR #485, opened 2017-04-18
Start of SyntheticCategories

jrochkind:
This is not the final UI. We don't have "splash pages" yet. I don't intend to expose
the synthetic categories on the normal search results page, like it is now. But this gives us
a smaller piece to review, including in production, to make sure the general idea is sound.

refs #434, #435, #436, #437, #438, #439

### DIFF
--- a/app/models/search_builder.rb
+++ b/app/models/search_builder.rb
@@ -12,5 +12,7 @@ class SearchBuilder < Blacklight::SearchBuilder
 
   include Hydra::AccessControlsEnforcement
   include CurationConcerns::SearchFilters
+
   include SearchBuilder::RestrictAdminSearchFields
+  include SearchBuilder::SyntheticCategoryLimit
 end
--- a/app/models/search_builder/synthetic_category_limit.rb
+++ b/app/models/search_builder/synthetic_category_limit.rb
@@ -17,9 +17,9 @@ class SearchBuilder
   #
   # Now when you do things with that search builder, it's locked to :portraits_and_people
   #
-  #     synthetic_category_param : set to a string, and this will be taken
-  #       from app params to identify a collection to limit to.  If unknown,
-  #       will do nothing. (Should limit to 0 rows instead?)
+  #     synthetic_category_param : set to a symbol, and this will be taken
+  #       from app params to identify a collection to limit to.  If a string
+  #        not matching a category is provided, will limit to 0 results, beware!
   #     synthetic_category_force: symbol of synthetic collection key,
   #       will just force limiting to there.
   module SyntheticCategoryLimit
--- a/app/models/search_builder/synthetic_category_limit.rb
+++ b/app/models/search_builder/synthetic_category_limit.rb
@@ -29,6 +29,11 @@ class SearchBuilder
       if synthetic_category
         solr_params[:fq] ||= []
         solr_params[:fq] << synthetic_category.solr_fq
+      elsif synthetic_category_param.present?
+        # trying to use param mapping for a category that doesn't exist, let's give you zero
+        # results!
+        solr_params[:fq] ||= []
+        solr_params[:fq] << "-*:*"
       end
 
     end
--- a/app/models/search_builder/synthetic_category_limit.rb
+++ b/app/models/search_builder/synthetic_category_limit.rb
@@ -1,0 +1,37 @@
+class SearchBuilder
+  # A search builder module to restrict to a certain SyntheticCollection.
+  # Won't do _anything_ just by being in the chain -- sets class_attributes
+  # on the SearchBuilder that can be used to turn it 'on', either on the class
+  # or even on an instance.
+  #
+  #.    synthetic_category_param : set to a string, and this will be taken
+  #.      from app params to identify a collection to limit to.  If unknown,
+  #.      will do nothing. (Should limit to 0 rows instead?)
+  #.    synthetic_category_force: symbol of synthetic collection key,
+  #.      will just force limiting to there.
+  module SyntheticCategoryLimit
+    extend ActiveSupport::Concern
+
+    included do
+      self.default_processor_chain += [:synthetic_collection_limit]
+      class_attribute :synthetic_category_param
+      class_attribute :synthetic_category_force
+    end
+
+
+    def synthetic_collection_limit(solr_params)
+      synthetic_category = if synthetic_category_force.present?
+        CHF::SyntheticCategory.new(synthetic_category_force)
+      elsif synthetic_category_param.present? && CHF::SyntheticCategory.has_key?(blacklight_params[synthetic_category_param])
+        CHF::SyntheticCategory.new(blacklight_params[synthetic_category_param])
+      end
+
+      if synthetic_category
+        solr_params[:fq] ||= []
+        solr_params[:fq] << synthetic_category.solr_fq
+      end
+
+    end
+
+  end
+end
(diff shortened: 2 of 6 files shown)


config/initializers/sufia_catalog_search_builder_overrides.rb
```diff
@@ -18,4 +18,9 @@
     klass.send(:include, SearchBuilder::RestrictAdminSearchFields)
   end
 
+  unless klass.ancestors.include? SearchBuilder::SyntheticCategoryLimit
+    klass.send(:include, SearchBuilder::SyntheticCategoryLimit)
+    klass.synthetic_category_param = :temp_synthetic_collection
```
hackmastera: I don't see this in the include above in app/models/search_builder. So I think that means we only want this behavior in the catalog search builder. Then I'm not sure why the module is included in app/models/search_builder at all.

app/models/search_builder.rb
```diff
@@ -12,5 +12,7 @@ class SearchBuilder < Blacklight::SearchBuilder
```
hackmastera: Since I'm asking questions, what are the non-catalog search builders? Are they used in 'My works' and similar areas (my only guess, really)? Or how can I find out what they are?

app/models/search_builder/synthetic_category_limit.rb
```diff
@@ -0,0 +1,37 @@
+class SearchBuilder
+  # A search builder module to restrict to a certain SyntheticCollection.
+  # Won't do _anything_ just by being in the chain -- sets class_attributes
+  # on the SearchBuilder that can be used to turn it 'on', either on the class
+  # or even on an instance.
+  #
+  #.    synthetic_category_param : set to a string, and this will be taken
+  #.      from app params to identify a collection to limit to.  If unknown,
+  #.      will do nothing. (Should limit to 0 rows instead?)
+  #.    synthetic_category_force: symbol of synthetic collection key,
```
hackmastera: Is this useful in testing somehow? Don't really understand; maybe you could give a use case or example.

app/models/search_builder/synthetic_category_limit.rb
```diff
@@ -0,0 +1,37 @@
+class SearchBuilder
+  # A search builder module to restrict to a certain SyntheticCollection.
+  # Won't do _anything_ just by being in the chain -- sets class_attributes
+  # on the SearchBuilder that can be used to turn it 'on', either on the class
+  # or even on an instance.
+  #
+  #.    synthetic_category_param : set to a string, and this will be taken
+  #.      from app params to identify a collection to limit to.  If unknown,
+  #.      will do nothing. (Should limit to 0 rows instead?)
```
jrochkind: The question is "what does this module do, if the code asks it for a synthetic category that does not exist."  I actually don't plan to make that _possible_ in how I ultimately use this (another part of code is going to end up aborting things before it even gets there), but was just thinking through all the possibilities. 

If you do have this module engaged, and ask for a synthetic category that doesn't exist... I guess getting 0 hits back is maybe what would be expected to happen? So maybe I should write that code anyway to be conservative, even though I don't plan for it to be possible in what I ultimately do.

app/models/search_builder/synthetic_category_limit.rb
jrochkind: The idea is that this module gets added to the search builder, and does _nothing_ unless it's activated. There are two ways to activate it:

1. Tell it to automatically take a param from the actual app query params, and map it to a synthetic category. 

2. Tell it to _always_ limit to a certain synthetic category. 

Both of these can be done either at the search builder _class_ level, or at the search builder _instance_ level.  The second one would be more likely at the instance level:

~~~
search_builder = # wherever you get it from
search_builder.synthetic_category_force = :portraits_and_people
search_builder.search # now when you do things with it, it's locked to portraits and people
~~~

It's the second one I actually plan to use in the "landing pages" for collections. This isn't how Hydra generally uses search builders.... but I think it probably should be, the search builder architecture is begging for it, it's very clean. 

Make sense? Maybe I need better inline docs here!

config/initializers/sufia_catalog_search_builder_overrides.rb
jrochkind: The idea is that the module is included, but inactive waiting to be activated, everywhere. But then it can be activated with the params discussed above. 

Hydra makes it confusing when the `::SearchBuilder` is in use, and when other search builders are in use. I thought it made sense to include the module "everywhere", so when you needed it you could just activate it without having to figure out "what search builder is in use now, and was the module included in it?"  Cause it's confusing. 

In general, when I'm adding things to the special catalog search builder through this weird hack, I'm also adding the to `::SearchBuilder` -- jcoyne has a PR in to make `::SearchBuilder` be the ancestor of _everything_, if that happens, we can get rid of the weird override hack and just use `::SearchBuilder`. 

See https://github.com/projecthydra-labs/hyrax/issues/707
https://github.com/projecthydra-labs/hyrax/pull/769

app/models/search_builder.rb
jrochkind: It's not just you. This is super confusing, and I think the Sufia/Hyrax architecture itself is confused, due to different people working on it at different times with different ideas. I don't completely understand it, and think there may be no 'simple' explanation, the code is weird and different code does different things. 

I brought this up in a Hyrax issue, see  projecthydra-labs/hyrax#707
projecthydra-labs/hyrax#769

app/models/search_builder/synthetic_category_limit.rb
hackmastera: Yes make sense, thanks! I think adding this into the comment as an example use case would be great.

app/models/search_builder/synthetic_category_limit.rb
hackmastera: If the alternative is "don't filter at all" then I think yes 0 hits makes more sense. I see your point about that won't be possible, though; I did notice the ArgumentError for that case. So I'm fine to merge as is.

jrochkind: Oh, I think it's possible probably, I'll see if I can do that. Don't expect to run into that case in actual use, but if it's easy to do, then probably makes sense 'just in case' of accidentally doing it.

jrochkind:
done.

app/models/search_builder/synthetic_category_limit.rb
```diff
@@ -0,0 +1,55 @@
...
+  #      SearchBuilder.synthetic_category_param = :category
+  #
+  # synthetic_category_force forces to a specific category, you might
+  # set it on a particular search_builder instance:
+  #
+  #      some_search_builder.tap { |sb| sb.synthetic_category_force = :portraits_and_people }.search(...)
+  #
+  # Now when you do things with that search builder, it's locked to :portraits_and_people
+  #
+  #     synthetic_category_param : set to a string, and this will be taken
```
hackmastera: This says "set to a string" but the example above uses a symbol.

app/models/search_builder/synthetic_category_limit.rb
```diff
@@ -0,0 +1,55 @@
...
+  #       from app params to identify a collection to limit to.  If unknown,
+  #       will do nothing. (Should limit to 0 rows instead?)
```
hackmastera: This comment needs to be updated to match new behavior

PR #488, opened 2017-04-19
Move generic_work translations to new generic_work locale file, fix #458

### LINKED ISSUE #458: Change Hydra tooltips back to local versions
Noticed that some of the hydra tooltips have reverted back to the default text. Happening in fields:

title,
external ID
language
description (now called abstract or summary again)
subject
related URL
rights (maybe?)

hackmastera:
fixes #458

The new file is broken out in support of per-work-type translations (e.g. if we want custom work types with custom metadata profiles for archives vs. oral histories). It was generated in the sufia 7.3 upgrade and I failed to fully attend to it.

Not sure what to make of the old one now; it just has one translation in it but I guess that's fine?

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -21,43 +21,6 @@
 
 en:
   simple_form:
-    labels:
-      generic_work:
-        genre_string: "Genre"
-        identifier: "External ID"
-        date_of_work: "Date"
-        description: "Description"
-        resource_type: "Format"
-        division: "Department"
     hints:
       defaults:
         files: '' # no file hint for now
-      generic_work:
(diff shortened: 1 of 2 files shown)


jrochkind:
Cool. FYI, the way you organize these in files is completely up to you/us, and doesn't effect I18n lookup. All that matters is the keys, you can split them between different files or keep them all in one file, with no semantic effect on the program. 

Splitting them between different files is just to help the developers keep track of things and do maintenance. So any way you want to do it is fine as far as the running program is concerned.

jrochkind:
That is, the way we break things into files with overrides doesn't need to match the way the upstream gems do either! Although it might be helpful for them to, for us figuring out how our keys relate to the upstream gems i18n keys they may be overriding.

Issue #466, opened 2017-04-13
Fix and expand integration tests


hackmastera:
We'll want to do this before we upgrade to hyrax, but maybe after getting on Rails 5; @jrochkind notes that "theoretically, Rails5 officially supports capybara integration tests now.... I'm skeptical how well it will work, but if we wait til rails5 for integration test bonanza, i'd probably want to try the official rails architecture at least."

in the meantime let's at least keep track of integration tests we want to write but haven't gotten to:
- [x] searchbuilder functionality in `config/initializers/sufia_catalog_search_builder_overrides.rb`
- [x] collection pages
- [x] synthetic categories page
- [x] end-to-end metadata form tests? especially permissions and visibility. Look into whether this is the right way to test this stuff, or we just need to expand unit tests we already have.
- [x] A basic happy path search. Enter a query, get results, click on a result. As both logged in and not logged in user.

hackmastera:
refs #706

MDiMeo:
When doing this, consider the manual workflows written for the QA process:

End User: https://chemheritage.atlassian.net/wiki/spaces/HDC/pages/129892353/QA+Workflow+-+EndUser

Admin Tasks: https://chemheritage.atlassian.net/wiki/spaces/HDC/pages/129925129/QA+Workflow+-+Admin+Tasks

jrochkind:
If we end up needing fits on travis, see: https://github.com/ucsblibrary/alexandria/blob/master/.travis.yml#L24-L27

jrochkind:
We're not going to have everything in manual QA workflows for now, but enough to make me a lot more confident in CI verifying our app still works for basic use cases, especially end-user use cases.

jrochkind:
merged to rebrand branch

PR #490, opened 2017-04-20
Synthetic Cateogry landing pages

jrochkind:
With more control of titles of synth categories. 

The `synthetic_category_param` is unset, no longer allow query params to
set category limits on ordinary results page, instead you use the category
landing pages, woot.

### DIFF
--- a/app/controllers/synthetic_category_controller.rb
+++ b/app/controllers/synthetic_category_controller.rb
@@ -1,0 +1,51 @@
+# Only a 'show' action
+#
+# Based roughly off of Sufia/CurationConcerns CollectionController#show
+# TODO link to CollectionControllerBehavior in both
+class SyntheticCategoryController < ApplicationController
+    # include Blacklight::AccessControls::Catalog
+    include Blacklight::Base
+
+    # ??
+    copy_blacklight_config_from(::CatalogController)
+
+    def show
+      unless synthetic_category
+        raise ActionController::RoutingError.new("No SyntheticCategory matches `#{params[:id]}`")
+      end
+      set_response # trigger setting of @response, which I think some partials need.
+    end
+
+
+    protected
+
+    # Get controller to find templates in CollectionController too,
+    # so we can re-use more of the sufia stuff for collection. Messy messy.
+    def self.local_prefixes
+      @local_prefixes ||= super.push('collections')
+    end
+
+    def synthetic_category
+      @synthetic_cateogry = if CHF::SyntheticCategory.has_key?(params[:id])
+        CHF::SyntheticCategory.new(params[:id])
+      else
+        nil
+      end
+    end
+    helper_method :synthetic_category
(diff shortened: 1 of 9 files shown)


lib/chf/synthetic_category.rb
```diff
@@ -82,15 +91,58 @@ def self.keys
...
+        self.new(slug.underscore)
+      else
+        nil
+      end
+    end
+
+
     attr_accessor :category_key
 
+    def slug
```
hackmastera: Can you add a comment explaining why this is necessary?

app/controllers/synthetic_category_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+      end
+      set_response # trigger setting of @response, which I think some partials need.
+    end
+
+
+    protected
+
+    # Get controller to find templates in CollectionController too,
+    # so we can re-use more of the sufia stuff for collection. Messy messy.
+    def self.local_prefixes
```
hackmastera: Are there upgrade scenarios wherein this would break? If so is there a test we could add to cover that?

lib/chf/synthetic_category.rb
jrochkind: Hmm....  I decided it was better for SEO and general look of professionalism to have the url look like `/focus/portraits-and-people` instead of `/focus/portraits_and_people`.  But it's difficult to work with symbols with dashes in them in ruby source code. So I don't know if it's truly _neccesary_, it just seemed convenient to me. What do you think?

app/controllers/synthetic_category_controller.rb
jrochkind: I guess a future version of Rails might break it. I'm not sure how to test for it without kind of copying Rails tests. 

Hmm, I guess I could write an integration/capybara test for this synthetic category page. I guess that'd be the right way to do it, that'd fail if the view template lookup stuff broke. And might be the answer to your question above "I am wondering if there are tests it would be useful to add here". 

What do you think, is it finally time to break through our integration test reluctance, and get one written for this, as a pre-req to it getting merged?  I'm not really sure.

lib/chf/synthetic_category.rb
hackmastera: Yes, I agree that the dashes are much better; makes sense.

app/controllers/synthetic_category_controller.rb
hackmastera: Here, I added it to this issue: https://github.com/chemheritage/chf-sufia/issues/466

jrochkind: So if we're okay postponing the integration test... ready to merge?  I just need to add a comment about the `slug`?

hackmastera: yes please.

PR #499, opened 2017-04-25
Add rake tasks to set up admin role, admin users

### LINKED ISSUE #448: create admin accounts
before admins can do stuff we need to be able to create them.

hackmastera:
fix #448

### DIFF
--- a/lib/chf/utils/admin.rb
+++ b/lib/chf/utils/admin.rb
@@ -1,0 +1,24 @@
+module CHF
+  module Utils
+    class Admin
+
+      def self.grant(email)
+        begin
+          u = User.find_by!(email: email)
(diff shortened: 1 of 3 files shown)


lib/chf/utils/admin.rb
```diff
@@ -0,0 +1,24 @@
...
+
+      def self.grant(email)
+        begin
+          u = User.find_by!(email: email)
+        rescue ActiveRecord::RecordNotFound
+          raise
+        end
+        r = Role.find_or_create_by!(name: 'admin')
+        u.roles.push(r) unless u.roles.include?(r)
+        r.reload
```
jrochkind: what makes the reload necessary/desirable?

PR #501, opened 2017-04-25
Add item count to synthetic category collection-like display.

jrochkind:
Unlike real collection display, it will show you the count of your result set when
you do a search-inside-the-thing too.

Put it in the space occupied by 'view type' selector in real collections, which we don't use,
since it was available space, and anna said she thought the blank space made things look odd,
and right under the search box seems like a good place when it updates for your current result set.

That means it goes in the "sort and per page" partial, even though it's not really about
sort and per page (neither are view types really!), in order to take advantage of existing
collection CSS consistently.


jrochkind:
screen shot 2017-04-25 at 12 12 33 pm

PR #504, opened 2017-04-27
Admin access

### LINKED ISSUE #478: admin should be able to view / edit any object

hackmastera:
refactor ability spec and demonstrate that admins already have update access to all files. closes #478

### DIFF
--- a/spec/models/ability_spec.rb
+++ b/spec/models/ability_spec.rb
@@ -2,42 +2,44 @@ require 'rails_helper'
 require 'cancan/matchers'
 
 RSpec.describe Ability do
+  let(:staff) { FactoryGirl.create(:user) }
+  let(:admin) { FactoryGirl.create(:admin) }
+  let(:admin_role) { Role.create name: "admin" }
+  before do
+    admin_role.users << admin
+    admin_role.save
+  end
 
   describe "an unprivileged user" do
-    let(:user) { FactoryGirl.create(:user) }
-
     it "is not an admin" do
-      expect(user.admin?).not_to eq true
+      expect(staff.admin?).not_to eq true
     end
 
-    subject { user }
-    it { is_expected.not_to be_able_to(:read, Role) }
-    it { is_expected.not_to be_able_to(:add_user, Role) }
-    it { is_expected.not_to be_able_to(:remove_user, Role) }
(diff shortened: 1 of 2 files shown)


jrochkind:
This is actually only changes to the _spec_, all that stuff was already working, we just had to verify it in a spec? Great!

spec/models/ability_spec.rb
```diff
@@ -2,42 +2,48 @@
 require 'cancan/matchers'
 
 RSpec.describe Ability do
+  let(:staff) { FactoryGirl.create(:user) }
+  let(:admin) { FactoryGirl.create(:admin) }
+  let(:admin_role) { Role.create name: "admin" }
+  before do
+    admin_role.users << admin
+    admin_role.save
+  end
```
jrochkind: If FactoryGirl has a factory defined `:admin`, shouldn't it already be doing this? I see in fact the only difference in the admin factory is the email! 

I think the admin factory should probably be fixed to just do this in the factory. Probably with a find_or_create so it's nice and lazy. 

Otherwise, I'd get rid of the admin factory altogether. And make the user factory use FG's sequence feature, so every one doesn't have the same email, and you can create multiple. 

What do you think?  Just suggestions, we can merge as is depending on your thoughts.

hackmastera:
@jrochkind better? Or were you thinking a factory for the admin role, also?

jrochkind:
@hackmastera yep, that's exactly what I was thinking! 

Except we probably want all the user factories to use factorygirl sequence feature for email addresses, so it will be possible to create more than one :admin or :user without their emails conflicting and raising a unique violation error (assuming we ensure emails are unique, which hopefully the code does!)

That doesn't seem to be a problem in this test or any existing tests, so up to you whether you want to fix it now (or ask me to :) ), but it's a good practice that will avoid an annoying problem the first time we create a test that does `create(:user)` more than once or whatever.

hackmastera:
@jrochkind How about this? A couple looked unused so I just removed them. I changed admin to a trait and got rid of its custom email address. `depositor` is in use and I don't know that I want to spend time messing with that one now; maybe we should just leave it. If this seems good to you and the tests pass I'm inclined to be done for now. Thanks for these suggestions to improve the specs!

hackmastera:
(the regular `user` factory is already using sequences)

PR #505, opened 2017-04-27
Only an admin can delete things

### LINKED ISSUE #267: only an admin should be able to delete objects

hackmastera:
closes #267

### DIFF
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -8,12 +8,6 @@ class Ability
 
   # Define any customized permissions here.
   def custom_permissions
-    # Limits deleting objects to a the admin user
-    #
-    # if current_user.admin?
-    #   can [:destroy], ActiveFedora::Base
-    # end
-
     if current_user.admin?
       # Role management
       # don't allow :destroy, :edit, :create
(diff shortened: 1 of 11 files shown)


jrochkind:
I don't really understand the code in Ability; do you know much of that is standard cancancan and how much is Hydra?  I've been reading up on cancancan (haven't used it before), but still getting confused. 

This part in particular is mystifying me:

~~~
    can [:edit, :update], String do |id|
       test_edit(id)
     end
~~~

I probably need to go look at the hydra_access_controls file it's updating and play with it to see how it works. 

Thanks for citing the file overridden in the comment! I like to put in a full github URL, and press `y` on github web to get a URL to the _specific version_ -- that way not only is it convenient, but you have a link to the actual version you were looking at when you overrode, so a later developer can easily see what's changed upstream since the override, if anything.

hackmastera:
Didn't know about the `y` shortcut in gh; that's cool. I went ahead and used the tag url since that gives even a bit more context.

expanded on the String permissions comment; I'll go ahead and merge as you suggested on slack.

PR #507, opened 2017-04-28
Add default workflow / admin set instructions to README

hackmastera:


jrochkind:
Rails apps have a `./bin/setup`  especially used for setting up a new dev environment. Runs `bundle install`, `db:create` (and schema load), and you are meant to add any custom things that need to be run... like these new ones. 

We might want to consider that? I'll play with it.

jrochkind:
Do you know if `rake curation_concerns:workflow:load`
and  `rake sufia:default_admin_set:create` are idempotent?

hackmastera:
Yes both are idempotent.

Never heard of `./bin/setup` that sounds ideal. Currently I'm running both in capistrano, but that would probably be better if it would be picked up by dev environments, also!

jrochkind:
Yeah `./bin/setup` wipes out the database, so can't be run in production, unless you're setting up a new machine.  So probably needs to be in both places, I guess -- if you also want to run it on deploy of a new version (in case workflows have changed I guess), makes sense for it to be in capistrano too I think.

PR #525, opened 2017-05-11
Don't return nil thumbnail_id, fix #517

### LINKED ISSUE #517: Works without a thumbnail cannot be viewed
When a work is made with metadata prior to the object being ingested it is now unable to be viewed. This is probably linked to the lightbox fix pushed out in (https://github.com/chemheritage/chf-sufia/commit/6f458d4d0db912d2d1d05267de97e3e6e40f4d3f).

Below is a copy of Sufia's error message in the log
```F, [2017-05-08T14:45:06.152062 #23646] FATAL -- : 
ActionView::Template::Error (No route matches {:action=>"show", :controller=>"downloads", :file=>"jpeg", :id=>nil} missing required keys: [:id])
:
    2:   <div class="modal-dialog">
    3:   <div class="modal-content">
    4:     <div class="modal-body">
    5:        <%= image_tag main_app.download_path(preview_id(presenter), file: "jpeg") %>
    6:      </div>
    7:      <div class="modal-footer">
    8:        <p><%= presenter.title.first %></p>
  app/views/curation_concerns/base/_lightbox_preview.html.erb:5:in `_app_views_curation_concerns_base__lightbox_preview_html_erb___108323443018272842_87992580'
  app/views/curation_concerns/base/show.html.erb:40:in `block in _app_views_curation_concerns_base_show_html_erb__2521692795715097227_87746420'
  app/views/curation_concerns/base/show.html.erb:39:in `each'
  app/views/curation_concerns/base/show.html.erb:39:in `_app_views_curation_concerns_base_show_html_erb__2521692795715097227_87746420'
```

hackmastera:


app/helpers/application_helper.rb
```diff
@@ -9,6 +9,6 @@ def license_label(id)
 
   # A work uses its thumbnail for preview; file just uses itself.
   def preview_id(presenter)
-    presenter.respond_to?(:thumbnail_id) ? presenter.thumbnail_id : presenter.id
+    presenter.respond_to?(:thumbnail_id) and presenter.thumbnail_id ? presenter.thumbnail_id : presenter.id
```
jrochkind: why `and` instead of `&&`? Avoid the need for parens?  I'd rather use the standard `&&` and parens and have it be clear what's going on (I had to stare at this a while and compare it to previous line to get it), but I approve this PR anyway, feel free to merge.

hackmastera: Oh, I have a couple areas where python style has stuck on me. I appreciate these things being pointed out so I can be more rubyish. I'll change it!

PR #543, opened 2017-05-22
Initial html outline for new homepage design

hackmastera:


jrochkind:
@hackmastera You're suggesting merging this into master now as an outline?  Works for me, just wanted to double-check.

hackmastera:
Yes because it has the added benefit of cleaning up the front page and making it easier to bear while we fill in the design.

app/views/layouts/homepage.html.erb
```diff
@@ -0,0 +1 @@
+<%= render template: 'layouts/sufia' %>
```
jrochkind: oh yeah, I forgot how to do this nested layouts in Rails, nice one.

jrochkind: oops, actually, not sure this will work. layouts/sufia has the wrapping `<html><body>` etc tags... which need to be on the outside, not wrapped by other content. 

May need to copy the layouts/sufia locally. :(  Ideally, the layouts/sufia might use partial(s) that we can reuse... but it doesn't. We could PR that to hyrax.

jrochkind: actually it does have some partials already. may be reasonable to copy/paste/alter it locally.

PR #548, opened 2017-05-23
Initial implementation of new design for collections index

hackmastera:
TODO:
- With current sizes there is no way to ever get 4 columns. Think about
  ways to allow 4 columns; maybe set container width proportionally?
- Justify center (currently in use) makes the elements in the last row come out of their
  columns. Justify left makes the entire flex area shift to the left of
  the page. Is there a way to make flexbox do what we want here or do we
  really need a grid?

Note a couple of cleanup commits also in this PR. I can split this up if desired.

### DIFF
--- a/app/assets/stylesheets/chf-overrides.css.scss
+++ b/app/assets/stylesheets/chf-overrides.css.scss
@@ -1,9 +1,3 @@
-// front page overrides
-#home_header  {
-  background-image: image-url('molecules_that_matter.jpg') !important;
-  background-position: 50% 50% !important;
-}
-
 // overrides for maker field
 .form-group {
   .double-input {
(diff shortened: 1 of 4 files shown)


jrochkind:
> With current sizes there is no way to ever get 4 columns. Think about
ways to allow 4 columns; maybe set container width proportionally?

Did the mockup from 2010 have four columns? I think eventually we will want to increase max column size, but it might needs to wait for a site-wide redesign to avoid messing up other things. And at that point, the flexbox will let it magically become four columns?

> Justify center (currently in use) makes the elements in the last row come out of their
columns. Justify left makes the entire flex area shift to the left of
the page. Is there a way to make flexbox do what we want here or do we
really need a grid?

I'm going to play around with this and see if I can find a solution, might not get to it til tomorrow.

jrochkind:
Note to selves also deal with illegal nested `<html>` tags do to nested layouts potential issue before merge.

PR #553, opened 2017-05-24
[WIP} Collections index

hackmastera:


jrochkind:
Should we go ahead and merge this even without the custom images? Everything will just have the 'default' image. @MDiMeo the default image right now is a kind of screwy looking license-free "archive box" SVG I found. We might want our own 'default' image, for when collections don't have a custom thumb assigned, which will probably happen from time to time despite out best efforts.

hackmastera:
+1 to merging without custom images.

jrochkind:
@hackmastera cool. ready for merge then you think?

PR #556, opened 2017-05-25
Frontpage controller

hackmastera:
Use our own controller / views for front page, since sufia uses its 'homepage' layout for all static pages.
Just use plain 'sufia' layout on front page.
Override the 'homepage' layout so our static pages have a lot less stuff on them (just the search bar).


jrochkind:
Do we actually need our own controller for front-page?  Isn't overriding the 'homepage' layout to just defer to `sufia` layout already doing what we need here, as on other static pages? Why do we need our own controller?

hackmastera:
Overriding the 'homepage' layout impacts all static pages. We want to remove the searchbar from the top of our front page, but not from all the other static pages. So we can't simply override the 'homepage' layout to meet both use cases.

hackmastera:
So this is predicated on the assumption that we want a searchbar on every page, including the about / help / contact pages. Should probably be confirmed.

hackmastera:
(they currently have a searchbar; maintaining existing functionality this way)

jrochkind:
I get it. If all you want/need to do is change the layout, there should be a way to do that with the previous sufia/homepage controller, rather than making a new one that doesn't have any of it's functionality. But I don't know if it causes any problems to do it this way?

Do you think it's better to replace the controller with our own blank one, instead of just setting it's layout?  Or is the 'homepage' controller actually used for more than just the homepage?

hackmastera:
I don't think the homepage controller is used for more than just the homepage. But I also don't think we need any of its functionality.

I guess I thought it would help make it more clear that the homepage layout isn't related to the front page stuff if they had different names.

We could do it by overriding the sufia homepage controller instead, and making the layout change there. It would mean we didn't have to change routes.rb.

hackmastera:
Oh I see you want to reopen the controller and change the layout that way? I guess we could do that, too. It's not my instinct, I guess. For me this way is more readable because everything is in expected places. But I am flexible!

jrochkind:
Normally I would prefer to do that because you get to keep all existing functionality, just change the one thing you want to change. 

So if really all we wanted to do was change the layout, it would be better to surgically do that, instead of replace the whole controller with our own. 

On the other hand, I don't really trust sufia functionality to remain backwards compat anyway, and maybe we _do_ want to replace the whole controller if we're not using it's features, maybe that's a plus. 

I just wanted to bring up the choice, I entrust you to decide how you want to do it here. :) Fine with me to merge this one and the other one.

hackmastera:
closing; see #559

Issue #552, opened 2017-05-24
new navbar


hackmastera:

hackmastera:
About, Help, and Contact all appear to use the homepage layout. The homepage layout is actually used all over the place, basically for anything mildly peripheral to the app:
- https://github.com/projecthydra/sufia/blob/7.3-stable/app/controllers/static_controller.rb
- https://github.com/projecthydra/sufia/blob/7.3-stable/app/controllers/concerns/sufia/contact_form_controller_behavior.rb
- https://github.com/projecthydra/sufia/blob/7.3-stable/app/controllers/pages_controller.rb#L6

application layout doesn't seem to be used at all (after much clicking around with a debug message in there, I never hit the debug).

I don't think we need a new layout. We can customize the header with the `masthead` partial, the navbar (where the search box with the `

hackmastera:
I think the right thing to do is customize the homepage layout for use on the static pages (like about, etc)

Then create our own homepage controller (maybe call it frontpage to make it easier to distinguish from sufia and because that's what we call it), inherit from Sufia's homepage controller but change the layout back to the standard sufia layout.

Customize header via the `masthead` partial and footer via the `shared/footer` partial. I *think* we can ignore the navbar stuff on the frontpage and it just won't show up.

hackmastera:
Or maybe we don't want to inherit from sufia's homepage controller since we don't really want anything it has. Even its recent items functionality seems slightly different from what we would want (I think it may be too broad for our use)

jrochkind:
I don't understand why we'd need/want more than one layout. Don't we want the new branded navbar on _all_ pages?  Is there any reason we need a separate layout for just homepage, or just static pages, or just anything?  Is there any reason we need more than one layout?

I guess if sufia already _is_ for reasons we can't figure out... then I don't really know what to do.

jrochkind:
Looks like sufia supplies an 'admin', a 'homepage', and a 'sufia' layout, where 'admin' and 'homepage' render 'sufia' internally. 

Can we just change all the controllers that use the 'homepage' layout to just use the 'sufia' layout instead?

Or change all controllers everywhere to use a 'chf' layout, that calls out to 'sufia' the same way 'homepage' and 'admin' do?  

We should be able to change what layouts controllers use... probably. But I can't quite figure out what the best design here is either, not clear to me what/why sufia is doing in the first place.

hackmastera:
[...]

PR #557, opened 2017-05-25
Add new navbar and reorganize navigation elements, refs #552

hackmastera:
Still more to do on the navbar, but this is in a mergeable state as we continue making progress.

screen shot 2017-05-25 at 11 31 58 am

### DIFF
--- a/app/views/_controls.html.erb
+++ b/app/views/_controls.html.erb
@@ -1,3 +1,4 @@
+<%# overrides https://github.com/projecthydra/sufia/blob/v7.3.0/app/views/_controls.html.erb %>
 <nav id="searchbar" class="navbar navbar-default navbar-static-top">
   <div class="container-fluid">
     <div class="row">
--- a/app/views/_controls.html.erb
+++ b/app/views/_controls.html.erb
@@ -1,0 +1,9 @@
+<nav id="searchbar" class="navbar navbar-default navbar-static-top">
+  <div class="container-fluid">
+    <div class="row">
+      <div class="searchbar-right navbar-right col-sm-7 col-md-6">
+        <%= render partial: 'catalog/search_form' %>
+      </div>
+    </div>
+  </div>
+</nav><!-- /.navbar -->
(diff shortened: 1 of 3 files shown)


app/views/_controls.html.erb
```diff
@@ -0,0 +1,9 @@
+<nav id="searchbar" class="navbar navbar-default navbar-static-top">
+  <div class="container-fluid">
+    <div class="row">
+      <div class="searchbar-right navbar-right col-sm-7 col-md-6">
+        <%= render partial: 'catalog/search_form' %>
+      </div>
+    </div>
+  </div>
+</nav><!-- /.navbar -->
```
jrochkind: when overriding a partial from sufia/CC (I think this is?), what do you think about leaving a comment with path/version it was overridden from (github 'y' url works nicely) and why? Not sure which partials in this PR are ours, and which are overridden.

app/views/_masthead.html.erb
```diff
@@ -0,0 +1,37 @@
+<% if current_ability.current_user.staff? %>
+  <%= render '/shared/select_work_type_modal' if create_work_presenter.many? %>
+  <nav id="staff-navbar" class="navbar navbar-inverse navbar-static-top staff-navbar" role="navigation">
+    <div class="container-fluid">
+      <div class="collapse navbar-collapse" id="staff-navbar-collapse">
+        <%# TODO: make these collapse to hamburger on mobile? two hamburgers is weird but it takes up lots of space otherwise %>
```
jrochkind: i don't know if this is your comment or sufia's, but I'm not worried about admin-only stuff on mobile, admin won't be working on mobile much/at all I think.

app/views/_controls.html.erb
hackmastera: Yes will do that thanks for the reminder. both of these are overrides.

app/views/_masthead.html.erb
hackmastera: my comment. I'm fine with this approach / assumption!

jrochkind:
okay, I'm not _totally_ sure about all this stuff, but let's go with it!

PR #558, opened 2017-05-25
Collections index thumbs

### LINKED ISSUE #550: custom crop images for each collection, for collections page
Related to #548

@hackmastera and I decided actually adding a FileSet object to Collections is the least terrible option for now, even though it means our custom crop collection thumbnails will be in fedora, where we dont' really need them. 

It will allow admins to upload these collection thumbs.  

We will eventually be able to easily use them for collection pages and collections in search results too, instead of that line-drawing-blocks, but that prob won't be done quite yet.

jrochkind:
A PR into the collections-index branch at #553.  Closes #550. 

Collection thumbs are just in the source repo under app/assets/images/collections. 

Once deployed, a collection can use an image by setting it's `represensative_image_path` to eg `toy_dummy_collection.jpg`.  A rake task is provided to let you do that conveniently with `id` to path/filename pairs:

     rake chf:collection_images[5h73pw048=toy_dummy_collection.jpg,9p2909328=whatever.jpg]

This is a kind of hacky design, but it was fairly easy, and will get us going. And works reasonably fine in dev/test. 

A default collection thumb is provided, just some random license-free svg 'archive box' art I found. I assume we wont' stick with this, but we need some default image. (We could make an image of the blocks if we really want to, although the blocks are actually a font-awesome glyph). 

I also updated the collection#show page to use this representative image, cause it was pretty easy.  

Couldn't figure out how to update the search results thumb for a collection in a reasonable way (found the partial, but couldn't figure out the presenter), so didn't do that for now.


hackmastera:
Looks good! Here's a new issue for the search results thumbnail (seems lower priority, but probably something we'll eventually want)

PR #561, opened 2017-05-30
Update navbar to match style documents, closes #552

### LINKED ISSUE #552: new navbar

hackmastera:


config/locales/blacklight.en.yml
```diff
@@ -1,6 +1,6 @@
 en:
   blacklight:
-    application_name: 'CHF Hydra'
+    application_name: 'Digital Collections'
```
jrochkind: Do we maybe want this to say "CHF Digital Collections"? It shows up in places like email subjects and stuff, as well as HTML `<titles>` (and from that: bookmarks, facebook shares, etc). I suppose one could even argue for "Chemical Heritage Foundation Digital Collections", I dunno.

config/locales/sufia.en.yml
```diff
@@ -2,7 +2,7 @@ en:
   curation_concerns:
     institution_name:   "CHF"
   sufia:
-    product_name:       "CHF Digital Collections (Beta)"
+    product_name:       "Digital Collections"
```
jrochkind: Same, if above in i18n is changed, this should probably match, I don't really know where one is used vs the other, matching seems a good bet.

hackmastera: Looks like there's a helper that overrides application_name with the value of product_name.

hackmastera:
updated translations and logo partial; please re-review

jrochkind:
looks good to me, not sure about travis failure?

hackmastera:
oops, typo. thanks travis. will fix.

Issue #535, opened 2017-05-17
Design Discussion Document for Work "Show" page


jrochkind:
We need to redesign our work "show" page (what you get when you click on a work in search results or from a collection; or from google or facebook or a bookmark) to meet our understanding of user needs. While we can redesign it in the smallest ways possible or more significantly, _some_ redesign is needed to meet goals expressed in #312, which I'd also summarize as:

* Better interactive experience for dealing with multi-page works, including good ways to skim/browse/get an overview/skip around. 
* In-browser full-resolution zoom-in to our high-res scans. 
* Multiple download options in various formats
* Understanding 'extracted work' relation to parent work (I'm actually mostly punting on this one for now at this moment)

And:

* Maybe just some better all around UI/UX -- the visual prominence/hieararchy of things currently does not match their importance or frequency of use to the users. 

Because of the 'in-browser zoom' requirement, our page will need to involve a JS-based pan-and-zoom viewer, which is _probably_ tile-based (IIIF, DZI, etc), since our individual full-res images even encoded as JPG are, I _think_ too big to deliver to the browser effectively all at once. (3M and up I think, for full-res jpgs of our images, which I think is probably too large to just deliver in whole without tiling). 

When I consider possible designs from examples and our discussions, I think there are two major categories:  1) The JS "Viewer" embedded directly on the page you see as soon as you get to it; vs 2) An additional click on the initial page is needed to open the 'viewer'. 

Both have advantages and disadvantages. I think one key question, is: Are most users most of the time (possibly weighted by personas/user groups of higher priority for us) coming to the individual Work page (by click on search results, google, facebook, etc):

1. Wanting to immediately view and explore the scanned images/photos, or
2. Needing context beyond the images themselves (primarily the textual metadata we supply) to _identify_ whether this is the item they were looking for or an item of interest, or otherwise understand it's context properly before diving into the images. 

I'm going to show a few basic wireframe mock-ups of different possibilities in both these categories. The intention is to stay away from the individual micro-details of each, but focus on these general macro-approaches as alternatives, and start thinking about possibilities. Likewise, while I'm listing some brief implementation considerations for context, they are not meant to be the focus. 

It's also possible none of these are in the right direction, or that this document/exersize is not actually helpful to us -- I'm open to that! 

# A page where viewer is NOT on original page, but takes another click

[...]

PR #571, opened 2017-06-01
WIP: Work show page

jrochkind:
Redesign of work show page, per discussion at #535. 

* Does not have a viewer yet, just the show page. 
* Code is kinda messy in some places. Had to override sufia in some odd ways.  I wanna figure out a way to clean up the CSS at least. Other feedback very welcome. 
* Currently using the derivative(s) that were there, not quite the right resolution(s), defering that until we have riiif. 
* If you wanna take it for a spin, be sure to look at it with and without admin login, it differs somewhat. 
* Will likely squash before merging, these commits are kinda messy, especially at first when I was experimenting a lot.


config/initializers/curation_concerns.rb
```diff
@@ -9,3 +9,13 @@
   end
 end
 
+# We're actually overriding from the sufia-set Sufia::FileSetPresenter, to our
+# own. to_prepare seems necessary to make it stick in dev.
+Rails.application.config.to_prepare do
+  CurationConcerns::MemberPresenterFactory.file_presenter_class = CHF::FileSetPresenter
```
hackmastera: Can these go in an app/presenters/..._override.rb file now that we have that mechanism?

app/assets/stylesheets/local/show.scss
```diff
@@ -1,29 +1,165 @@
...
+    @media (min-width: $screen-lg-min) {
+      // 3 wide
+      width: calc(50% / 3 - 10px);
+      img.lazyload, img.lazyloading {
+        height: 15vw;
+      }
+    }
+  }
+
+  @media (min-width: $grid-float-breakpoint) {
```
hackmastera: curious why you use `grid-float-breakpoint` here instead of `screen-sm-min` since they are the same. Never really used the former. Have you seen cases where you need to change one to be different from the other?

app/assets/stylesheets/local/show.scss
```diff
@@ -1,29 +1,165 @@
...
+    width: 100%;
+    // keep crazy long aspect ratio images from being
+    // so long, although there will be some distortion.
+    max-height: 80vw;
+  }
+}
+
+.show-page-image-viewer-link {
+  display: block;
+  position: absolute;
```
hackmastera: can you comment the display / position settings here? Sometimes I find it helpful to have a little extra context on these since they are often about setting up behavior in child elements.

jrochkind:
> Reviewing CSS is hard. There are places where we could probably pull numbers into variables; I guess you'll take a pass for that?

figuring out how to write CSS in a clean way is something I find hard! I def think the CSS file needs a go-over with thoughts on refactoring/reordering to be clearer. I'll keep your numbers-into-variables suggestion in mind when I do that, thanks!

app/assets/stylesheets/local/show.scss
jrochkind: Nope, no legit reason, I just got confused, I think you are right!

config/initializers/curation_concerns.rb
jrochkind: Hmm... I think they can! Not sure if that makes sense for just setting a class variable... but if you think it does, I'm happy to roll with it!

hackmastera: To me it makes sense because it brings some consistency to class overrides if we try to do it the same way: class overrides live in `/app/, configuration lives in `config/`. I was only setting a variable when I set it up :) But it's also nice to have it that way already if/when we need to add more overrides to the class in question.

jrochkind: Work for me! 

I don't really think of just setting a global variable as an "override" or "monkey patch".  it's more like a configuration setting. It does need the `to_prepare` because of Rails dev-mode reloading though.  But I can see it going both ways, works for me.

jrochkind:
Here are some example docs on staging

one graphic (painting): https://staging.hydra.chemheritage.org/concern/generic_works/rj430479p
lots of pages: https://staging.hydra.chemheritage.org/concern/generic_works/tx31qj045
photo with one extra image: https://staging.hydra.chemheritage.org/concern/generic_works/w9505086g
single photo: https://staging.hydra.chemheritage.org/concern/generic_works/pn89d700k

ton of pages, some are child works, some are not public: https://staging.hydra.chemheritage.org/concern/generic_works/4b29b614k

jrochkind:
@hackmastera okay, I consider what's here done for now. 

Not sure if I'm going to do the viewer as a separate branch on top of this or just in here -- any opinions?  And either way, if you have any more feedback on what's here now, please provide it, or I'll consider it done (just not merged until viewer is ready).

hackmastera:
I think a separate branch for the viewer would be nice.

Are you still thinking of squashing these at all? I think you mentioned you might do that.

hackmastera:
tidied css looks good; thanks for the extra comments!

jrochkind:
> Are you still thinking of squashing these at all? I think you mentioned you might do that.

Not sure. The first few commits are definitely messy, but the last half or so document intent to some extent. What do you think, would you prefer I squash them all into one commit? I guess I've got to do that before I branch off it if I'm going to do it.

hackmastera:
re: squash I could go either way. We can always compare between shas if we  want to see it all at once, if you decide not to squash.

app/views/curation_concerns/base/_chf_show_page.html.erb
```diff
@@ -0,0 +1,4 @@
+<div class="work-show-page-thumb">
```
hackmastera: not sure I understand this partial. It doesn't seem like it's the representative image, judging from comment in `_show_page_image`. Maybe rename this to distinguish from that one? and/or add a comment explaining what it's about?

jrochkind:
Good comments, thanks. Is the lightbox stuff used in any of the admin pages? I wasn't sure if I could remove it safely.

app/views/curation_concerns/base/_chf_show_page.html.erb
jrochkind: I'm not sure either! Have to review the code and re-figure it out and comment or alter appropriately, thanks!

hackmastera:
I can take on a ticket to hunt for lightbox cleanup one this PR is merged, if that's easier

jrochkind:
> I can take on a ticket to hunt for lightbox cleanup one this PR is merged, if that's easier

Good thought. I don't mind doing it, but it might be easier to do it in a separate thing once this is merged, the complexity of my two PR's with one based on the other is already getting to me, ha.

hackmastera:
https://github.com/chemheritage/chf-sufia/issues/632

app/views/curation_concerns/base/_chf_show_page.html.erb
jrochkind: i believe the answer is that chf_show_page isn't actually used anymore and should be removed! good catch.

PR #574, opened 2017-06-05
Front page layout

### LINKED ISSUE #568: implement new front page design

hackmastera:
Will fix #568 after one last TODO: featured collection image needs some attention w/r/t responsive resizing.

Posted in the meantime for discussion of image treatment, especially resolution / retina display.


app/assets/stylesheets/local/index.scss
```diff
@@ -1,9 +1,147 @@
-.range_limit {
```
jrochkind: Is this stylesheet really just for the home page?  Can we call it `home_page.scss` or something instead of `index.scss`?  There are lots of 'index' pages.

hackmastera: So glad you asked. I meant to move out the couple of non-homepage things but never did and they ended up nested in the homepage wrapper. That's now fixed. Renamed the file to homepage.

PR #578, opened 2017-06-06
update capistrano and related gems

hackmastera:


Gemfile
```diff
@@ -55,10 +55,10 @@ group :production do
...
 
 group :development do
-  gem 'capistrano', '3.4.0'
-  gem 'capistrano-bundler', '1.1.4'
+  gem 'capistrano', '3.8.1'
+  gem 'capistrano-bundler', '1.2.0'
   gem 'capistrano-passenger', '0.2.0'
-  gem 'capistrano-rails', '1.1.3'
+  gem 'capistrano-rails', '1.2.3'
   gem 'capistrano-maintenance', '~> 1.0', require: false
```
jrochkind: Why are we locking to specific versions of cap in the gemfile?  I'd do more like `gem "capistrano", "~> 3.8", meaning any 3.x greater than or equal to 3.8.x.

hackmastera: It was locked specifically when I got to it, I think from when this was set up for us by DCE. I wasn't sure whether capistrano was following semantic versioning when I read over the release notes, so I didn't change it.

I can do it now, though.

PR #581, opened 2017-06-08
Fix flexbox display on safari (8.0.4)

hackmastera:


jrochkind:
weird, safari doesn't support flexbox without vendor prefix? caniuse.com says it's supposed to, and I had no troubles with flex in general. Whatever works, I guess!

PR #592, opened 2017-06-12
Hide search dropdown for non-staff users

### LINKED ISSUE #589: remove searchbox dropdown for non-staff

hackmastera:
fixes #589


jrochkind:
You're copying the entire search partial and then modifying?

Can you add a comment to the top explaining where it was copied from and what's been modified?

Also, i wonder if it would make sense to submit this upstream -- in cases where there is only one choice in that popup, don't display it?

hackmastera:
Sure, I added a comment at the top with a link. Note there's also a comment above the line I changed. Also I did this as 2 commits (add file w/o modifications first) so that blame / history will call out the changes.

Good point about fixing upstream. I think an upstream fix along the lines you suggest would probably be a different implementation, so maybe not something I will do right now. I will create a ticket, though, if nothing else.

PR #593, opened 2017-06-12
Correct searchbox element heights in ff / safari

### LINKED ISSUE #590: fix searchbox styles on FF / Safari
on hompage, the text area and button aren't lined up correctly

hackmastera:
fixes #590

### DIFF
--- a/app/assets/stylesheets/local/homepage.scss
+++ b/app/assets/stylesheets/local/homepage.scss
@@ -48,9 +48,9 @@
 
   .search-callouts {
     .input-group {
-      height: 46px;
+      min-height: 46px;
       input.q, button {
-        height: 46px;
+        min-height: 46px;
       }
     }
     @media (min-width: $screen-sm-min) and (max-width: $screen-sm-max) {
--- a/app/assets/stylesheets/local/homepage.scss
+++ b/app/assets/stylesheets/local/homepage.scss
@@ -50,7 +50,7 @@
     .input-group {
       height: 46px;
       input.q, button {
-        height: 100%
+        height: 46px;
       }
     }
     @media (min-width: $screen-sm-min) and (max-width: $screen-sm-max) {


jrochkind:
Interesting! Where does that 46px come from, is there any way to make it in terms of other existing bootstrap variables? Is there a danger of it changing if the default font size changes?

hackmastera:
the 46 comes from the containing element; I pulled the size off of the design. I'm not sure why setting the children to 100% didn't work.

Do you think I should try changing these to min-height?

jrochkind:
I think there's something built into bootstrap to do a button next to an input.... yup

http://getbootstrap.com/components/#input-groups

Does the standard bootstrap markup for this not work? Ideally they'd have figured out all the edge cases and browser-specific stuff, and we could just use the bootstrap markup and not need to add any of our own at all. Not working?

hackmastera:
it's in an input-group and it works fine until I try to make it bigger to get the chunkier look in the design. do you want to come look?

hackmastera:
oh i see input-group-lg; let me try that.

jrochkind:
Oh, I see you are using .input-group. Sorry, I'm behind!

It might be worth looking at how bootstrap itself handles differnetly sized inputs. They already have an `input-group-lg` etc. If those sizes aren't right, maybe define a new one the same way bootstrap does?

Which might be a hard-coded pixel height already, I dunno, heh. 

https://github.com/twbs/bootstrap-sass/blob/5d6b2ebba0c2a5885ce2f0e01e9218db3d3b5e47/assets/stylesheets/bootstrap/_input-groups.scss#L44-L53

https://github.com/twbs/bootstrap-sass/blob/5d6b2ebba0c2a5885ce2f0e01e9218db3d3b5e47/assets/stylesheets/bootstrap/_input-groups.scss#L44-L53

https://github.com/twbs/bootstrap-sass/blob/5d6b2ebba0c2a5885ce2f0e01e9218db3d3b5e47/assets/stylesheets/bootstrap/_forms.scss#L202-L205

Okay, sorry, you may have already done all this research, I see how your approach might make sense. 

Feel free to merge at will yourself!

hackmastera:
So the problem appears to be that we don't necessarily want the searchbox to be large everywhere -- just on the homepage. And the input-group classes are in a partial that is reused all over the site. So I don't think we can do that unless I add the class in javascript?

hackmastera:
will try it using $input-height-base scoped to just this page.

hackmastera:
Yeah I don't think you can do a sass variable that way, change it only within one class.

hackmastera:
Okay I tried a few other things and nothing else worked, really. Appreciate the ideas and urge/encouragement to find a better way. Merging as min-height.

jrochkind:
> So the problem appears to be that we don't necessarily want the searchbox to be large everywhere -- just on the homepage. And the input-group classes are in a partial that is reused all over the site. So I don't think we can do that unless I add the class in javascript?

You could potentially have made it apply just to a searchbox that is on the home page, with something like:

~~~
.home-page {
   .search-box {
      @extend .input-lg;
    }
}
~~~

Not sure if that will work or not. There's probably _some_ way to do it with selectors such that they just applying to the home page. 

But I think it's good enough!

jrochkind:
indeed sass variables should be considered globals that apply everywhere, you can't really change em just for a certain spot, or at any rate it's not a good idea to try. :)

PR #595, opened 2017-06-12
Place search dropdown (when visible) before submit button, round butt…

hackmastera:
…on corners


jrochkind:
This is more stuff the input-group is just supposed to take care of, and I wish we knew why it wasn't and why we're reimplementing bootstrap styles here, but merging!

hackmastera:
Actually I realized a better way to do this yesterday. The bootstrap style will take effect if the button is truly the last child. Submitting a new PR.

PR #607, opened 2017-06-14
Use the same blue on our buttons that we use in our header

hackmastera:
refs #496


hackmastera:
hm, actually not sure this is working right; close for now.

jrochkind:
Should we just be making that light blue our bootstrap $brand-primary?  That will make it be the default color for all links (which seems to be chemheritage.org website style), as well as button-primary's, etc. Would that make sense?

I didn't quite catch why you made it brand-secondary instead of brand-primary.

Issue #575, opened 2017-06-05
Confusion when a child work (or any filesets) is not public in a public work


hackmastera:
Broadening this to apply to "non-public" things in general on a public work:  Child works as well as ordinary fileset images; when they are the 'representative' image or just an ordinary member (this one is the hardest); on both show page and viewer. 

Probable solution: Either they need to not show up at all (for a non-logged in user who can not actually see them), or they need a more legible placeholder. 

The thumbnail appears in the list as a default image, with 'chf' in the top corner. Pretty confusing for a user. Maybe a different thumbnail? and/or a different message

jrochkind:
and/or consider it maybe not showing up at all to those who don't have access to see it. Although sounds like there were reasons that wouldn't work, worth putting in as another possible option in the ticket I think.

jrochkind:
@MDiMeo @catlu @archivistsarah 

If a not logged in user is viewing a show page, should non-public members just not show up at all?  That makes sense to me, I think?

archivistsarah:
yeah, that makes the most sense to me. unless there's an alternative I'm missing?

jrochkind:
I guess the alternative I can think of would be showing placeholders for those images like currently, but having a better UI for when you try to actually look at them, saying "no, you can't look at this cause it's not public" or something. I think not showing them at all is probably sensible.

jrochkind:
@hackmastera actually, can you clarify what you were pointing to here? Can you find or make an example on staging?  Actually looking at current behavior, i start to get confused. 

Is it specifically when a non-public fileset is listed as 'representative' on an otherwise public work, when viewed by a not logged in user?

jrochkind:
Sorry, I started talking about a different case above, then realized there were at least a couple different cases going on here.

hackmastera:
@jrochkind I created this during a team meeting. I think we were discussing the fact that there was a default thumbnail with the 'chf' flag, and there was nothing about that display that indicated the file was not actually viewable to the general user. If I recall, people in the room weren't sure about removing them altogether, because they worried users would notice / ask for the missing page. I think someone suggested some text like "coming soon" or "in process"? assuming it's still in cataloging. Could be a shaded box with brief text and no default thumbnail.

jrochkind:
OK, I guess I'll wait for more discussion before working on this further. 

Lots of things can be done, with varying levels of how long it will take to do, and varying levels of fighting with the stack. 

[...]

PR #618, opened 2017-06-15
Image viewer

jrochkind:
This is pulled on top of work_show_page branch. 

It's not _quite_ done, but close. 

UI to note:

* Still no high-res images, so you can't actually zoom in very far. 
* tab-to-link nav works on show page and viewer. (mostly, I think). This was surprisingly challenging, but worth it!
* Try out the "full screen" button especially on a large desktop display. How's that for interacting with a text?!?
* Keyboard shortcuts, click the keyboard button to see em
* Yes, I'm not totally happy with the bottom nav bar either, but haven't come up with something better, and I feel like running out of budgetted time on this, have to move on to something else, probably good enough for now. 
* There is a bug in header/footer layout on show page I still need to figure out. 

I am most interested in if you can find any _bugs_, things that don't work right, rather than feedback on colors/spacing/layout/etc.  The bugs are higher priority of course!

Technical things I need to look into:

- [x] Google analytics
- [x] Fix on IE11. (Postponing IE10 for now, decided it's okay it doesn't)
- [ ] Non public members should not show up for the public. See #575
- [x] bug where next doc right after opening the viewer doesn't update url
- [x] reorganize class and id names to be consistent, all start with `viewer-`. 
- [x] Escape on an open popup menu should close the menu not the viewer
- [x] There is a bug in header/footer layout on show page I still need to figure out. 
- [x] next/prev buttons not vertically centered on FF
- [x] need a loading spinner on page image change, especially once we're using slower riiif
- [x] next/prev scroll to thumb should be more gentle
- [x] remember focused image on show page keep it focused when you close viewer (better for keyboard nav)
- [x] for a child work with a long title, it pushes the rest of the viewer buttons off screen for me
for ex: https://staging.digital.chemheritage.org/concern/generic_works/h128nd97r/viewer/8c97kq68v
- [x] for the same work as above, when i click on the 2nd or 3rd (or any) child work, it always brings up the representative work instead, this doesn't happen for a work with only files
- [x] thumbs wrapping at one thumb instead of two in hard to repro way

### DIFF
--- a/app/assets/javascripts/chf_ga_events.js
+++ b/app/assets/javascripts/chf_ga_events.js
@@ -1,0 +1,14 @@
+// Sufia/hyrax JS for sending event tracking to Google Analytics is a bit inflexible.
+// https://github.com/samvera/hyrax/blob/1e504c200fd9c39120f514ac33cd42cd843de9fa/app/assets/javascripts/hyrax/ga_events.js
+//
+// We add our own more flexible.
+//
+// https://developers.google.com/analytics/devguides/collection/gajs/eventTrackerGuide
+
+$(document).on('click', '*[data-analytics-event]', function(e) {
+  _gaq.push(['_trackEvent',
+             e.target.getAttribute("data-analytics-event"),
+             e.target.getAttribute("data-analytics-label"),
+             e.target.getAttribute("data-analytics-value")
+            ]);
+});
--- a/app/assets/javascripts/chf_image_viewer.js
+++ b/app/assets/javascripts/chf_image_viewer.js
@@ -1,0 +1,497 @@
+// A JS 'class' for dealing with the image viewer
+
+function ChfImageViewer(element) {
+  this.element = element;
+  this.initModal(document.querySelector("#chf-image-viewer-modal"));
+}
+
+ChfImageViewer.prototype.viewerPathComponentRe = /\/viewer\/(\w+)$/;
+
+ChfImageViewer.prototype.show = function(id) {
+  if (document.activeElement && document.activeElement.getAttribute("data-trigger") == "chf_image_viewer") {
+    this.previousFocus = document.activeElement;
+  }
+
+  if (typeof this.viewer == "undefined") {
+    this.initOpenSeadragon();
+  }
+
+  this.totalCount = parseInt(document.querySelector(".viewer-thumbs[data-total-count]").getAttribute("data-total-count"));
+  if (this.totalCount == 1) {
+    // hide multi-item-relevant controls
+    document.querySelector("#viewer-pagination").style.display = "none";
+    document.querySelector("#viewer-right").style.display = "none";
+    document.querySelector("#viewer-left").style.display = "none";
+    document.querySelector("#viewer-thumbs").style.display = "none";
+  }
+
+  if (! OpenSeadragon.supportsFullScreen) {
+    document.querySelector("#viewer-fullscreen").style.display = "none";
+  }
+  if (! this.viewer.drawer.canRotate()) {
+    //OSD says no rotate
+    document.querySelector("#viewer-rotate-right").style.display = "none";
+  }
+
+  var selectedThumb;
+  // find the thumb
+  if (id) {
+    selectedThumb = document.querySelector(".viewer-thumb-img[data-member-id='" + id + "']");
+  }
+  if (! selectedThumb) {
+    // just use the first one
+    selectedThumb = document.querySelector(".viewer-thumb-img");
+  }
+  this.selectThumb(selectedThumb);
+
+  // show the viewer
+  $(this.modal).modal("show");
+  this.scrollSelectedIntoView();
+};
+
+// position can be 'start', 'end'
+ChfImageViewer.prototype.scrollSelectedIntoView = function(position) {
+  // only if the selected thing is not currently in scroll view, scroll
+  // it to be so.
+  // https://stackoverflow.com/a/16309126/307106
+
+  var elem = this.selectedThumb;
+
+  var container = $(".viewer-thumbs");
+
+  var contHeight = container.height();
+  var contTop = container.scrollTop();
+  var contBottom = contTop + contHeight ;
+
+  var contWidth = container.width();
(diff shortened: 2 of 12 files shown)


catlu:
Potential bugs:

- for a child work with a long title, it pushes the rest of the viewer buttons off screen for me
for ex: https://staging.digital.chemheritage.org/concern/generic_works/h128nd97r/viewer/8c97kq68v

- for the same work as above, when i click on the 2nd or 3rd (or any) child work, it always brings up the representative work instead, this doesn't happen for a work with only files

jrochkind:
It looks like most of our 2d matter is scanned against a _black_ background, the edges are black(ish).  

Would it make more sense to make the background of the image area black, to blend in with those edges?

jrochkind:
Bug to track down:

On some laptops, the thumbs are not wrapping properly in viewer, one column instead of two. Although I'm having difficulty reproducing. On personal laptop this one seemed not to be one column, although on work computer it is appropriately two column?

https://staging.digital.chemheritage.org/concern/generic_works/8p58pd02r/viewer/2n49t1788

catlu:
On my work computer, it switches from one column to two between my monitors--actually goes to one column on my wider screen (primary) monitor.

jrochkind:
@jrochkind Issue found, on Firefox 54, using the `Shifit`+`Up` or `Down` to zoom also causes the keyboard shortcut menu to appear and then you can't use the `Shift`+`Up` or `Down` until you close the menu. `+` and `-` work fine.

jrochkind:
@hackmastera This is basically done for now:

* not gonna worry about IE10 for now, @MDiMeo and communications agreed. 
* waiting on a decision about what to do with non-public members of public works
* Needs 'download' links, which I'm going to wait for riiif integration for. 
 
Do you want to review this as a separate PR before I squash it and merge it into #571? Up to you.

jrochkind:
Oops, spoke too soon, not currently working right in IE11 either! It used to be!

I'll get that fixed. But you are welcome to review at any time if you want.

hackmastera:
yes i prefer to review as a separate PR; will start looking at it now.

app/assets/javascripts/chf_image_viewer.js
```diff
@@ -0,0 +1,498 @@
...
+  var contRight = contLeft + contWidth;
+
+
+  var elemTop = $(elem).offset().top - container.offset().top;
+  var elemBottom = elemTop + $(elem).height();
+  var elemLeft = $(elem).offset().left - container.offset().left;
+  var elemRight = elemLeft + $(elem).width();
+
+  var isTotal = (elemTop >= 0 && elemBottom <= contHeight && elemLeft >= 0 && elemRight <= contWidth);
+  //var isPart = ((elemTop < 0 && elemBottom > 0 ) || (elemTop > 0 && elemTop <= container.height())) && partial ;
```
hackmastera: delete old code?

app/views/curation_concerns/base/_chf_image_viewer.html.erb
```diff
@@ -0,0 +1,63 @@
+<%# The image viewer, contained in a partial. Might later change this to
+    some kind of JS templates or other fancy JS, instead of just server-side
+    HTML all pre-rendered invisible on page like it is now.
+
+    local args:
+      * work
```
hackmastera: nice

app/assets/javascripts/chf_ga_events.js
```diff
@@ -0,0 +1,14 @@
+// Sufia/hyrax JS for sending event tracking to Google Analytics is a bit inflexible.
+// https://github.com/samvera/hyrax/blob/1e504c200fd9c39120f514ac33cd42cd843de9fa/app/assets/javascripts/hyrax/ga_events.js
```
jrochkind: Yep, it's not an override, it's separate JS. The one built into Hyrax (currently) looks for elements with a certain id, and assumes they have a certain GA event/label, and looks up the GA value from the 'label' tag. 

This one looks for elements with a data-analytics-event attribute, and gets all GA properties from data-* attributes. 

So, yeah, they'll apply to different elements, and both work, coexisting.  I thought it was worth establishing our own convention, cause the one built into sufia/hyrax is very inflexible. 

Theoretically this could be upstreamed to hyrax, but then you have to worry about breaking existing code, and I just don't really want to deal with it right now.

app/assets/stylesheets/local/chf_image_viewer.scss
```diff
@@ -0,0 +1,315 @@
...
+
+      font-size: $font-size-h3;
+
+      overflow: hidden;
+
+      a {
+        display: block;
+        white-space: nowrap;
+        overflow: hidden;
+        text-overflow: ellipsis;
```
jrochkind: it's kind of tricky to use, you need to make sure the element it's in has an explicit or implicit width or max-width, in order for there to be 'overflow' at all, when `white-space: nowrap`, in order for `text-overflow: ellipsis` to apply. 

It got tricky with all the flexboxes, and ended up needing some things in IE11 that weren't needed elsewhere. 

But it is a pattern I like, for dealing with unpredictable dynamic content that could be very long, where you want to restrict it to just what fits in one line in some views.

jrochkind: i also used it in child-work names on the show page thumbs, to restrict em to one line.

app/assets/javascripts/chf_ga_events.js
hackmastera: I think if we're not using the upstream behavior it'd be better to override so we can't accidentally use the id and start receiving weird / bad analytics data. Do you think there is a benefit to keeping the upstream code in play?

jrochkind: The benefit is if some other template from one of the upstream gems, now or in the future, assumes the upstream GA code is there, and we don't notice it's relying on it. I'm not sure, I see both sides. I prefer not overriding upstream code when I don't have to, as a general principle.

hackmastera: Okay, i'm not worked up about it; will trust your judgment!

PR #627, opened 2017-06-19
autoprefixer config file

jrochkind:


hackmastera:
Should we add a comment about our explicit support goals, vs. our use of the autoprefixer to mitigate awfulness?

jrochkind:
Sure, but I'm not sure our support goals are explicit yet. IE10, maybe, or else IE11?

hackmastera:
Pretty sure we agreed on 11. Perhaps @MDiMeo can confirm.

Issue #624, opened 2017-06-19
characterization is broken; reports wrong sizes


hackmastera:

hackmastera:
example:
https://digital.chemheritage.org/concern/parent/zc77sq40p/file_sets/hx11xf67b
actual dimensions are 2226 × 1650

hackmastera:
characterization actually returns both values

```
From: /home/ubuntu/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/hydra-works-0.16.0/lib/hydra/works/services/characterization_service.rb @ line 62:
Owner: Hydra::Works::CharacterizationService
Visibility: protected
Number of lines: 6

def parse_metadata(metadata)
  omdoc = parser_class.new
  binding.pry
  omdoc.ng_xml = Nokogiri::XML(metadata) if metadata.present?
  characterization_terms(omdoc)
end
[4] pry(#<Hydra::Works::CharacterizationService>)> characterization_terms(omdoc)
=> {:fits_version=>["0.8.5"],
 :format_label=>["Tagged Image File Format"],
 :file_mime_type=>["image/tiff"],
 :exif_tool_version=>["9.13"],
 :file_size=>["11082294"],
 :last_modified=>["2017:06:20 13:09:45-04:00"],
 :filename=>["2012-002_928.tif"],
 :original_checksum=>["b8344bf2f88bd421986d9ac988578e28"],
 :date_created=>["2016:08:22 11:40:49"],
 :well_formed=>["true"],
 :valid=>["true"],
 :byte_order=>["little endian", "little endian"],
 :compression=>["Uncompressed"],
 :width=>["2226", "160"],
 :height=>["1650", "119"],
 :color_space=>["RGB"],
 :profile_name=>["Adobe RGB (1998)"],
 :profile_version=>["2.1.0"],
 :orientation=>["normal*"],
 :capture_device=>["digital still camera"],
 :scanning_software=>["Capture One 8 Macintosh"],
 :exif_version=>["0220"]}

```

hackmastera:
It seems like we could fix this at access time for every access, but I think it would be better to change it here in hydra-works before it is stored, since the smaller dimensions are misleading.

hackmastera:
[...]

PR #629, opened 2017-06-21
Fix characterization

hackmastera:
refs #624; before the issue can be closed we need to run the new task

### DIFF
--- a/app/services/characterization_service_override.rb
+++ b/app/services/characterization_service_override.rb
@@ -3,15 +3,13 @@
 # are included for both 'layers' in the TIFF, we want to make sure we're saving
 # the one from the LARGEST layer, assuming that's the "real" layer, and ONLY
 # that one.
-module Hydra::Works
-  class CharacterizationService
-    protected
-      def append_property_value(property, value)
-        value = object.send(property) + [value] unless property == :mime_type
-        # We don't want multiple heights / widths, pick the max as the true
-        # width/height.
-        value = [object.send(property).to_a.concat(value).max] if property == :height or property == :width
-        object.send("#{property}=", value)
-      end
-  end
+Hydra::Works::CharacterizationService.class_eval do
+  protected
+    def append_property_value(property, value)
+      value = object.send(property) + [value] unless property == :mime_type
+      # We don't want multiple heights / widths, pick the max as the true
+      # width/height.
+      value = [object.send(property).to_a.concat(value).max] if property == :height or property == :width
+      object.send("#{property}=", value)
+    end
 end
--- a/app/services/characterization_service_override.rb
+++ b/app/services/characterization_service_override.rb
@@ -1,0 +1,13 @@
+# override https://github.com/samvera/hydra-works/blob/v0.16.0/lib/hydra/works/services/characterization_service.rb
+# to keep our multi-layer tiffs from causing problems
+module Hydra::Works
+  class CharacterizationService
+    protected
+      def append_property_value(property, value)
+        value = object.send(property) + [value] unless property == :mime_type
+        # We don't want multiple heights / widths, it doesn't make sense.
+        value = [object.send(property).to_a.concat(value).max] if property == :height or property == :width
+        object.send("#{property}=", value)
+      end
+  end
+end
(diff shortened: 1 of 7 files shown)


app/services/characterization_service_override.rb
```diff
@@ -0,0 +1,13 @@
+# override https://github.com/samvera/hydra-works/blob/v0.16.0/lib/hydra/works/services/characterization_service.rb
```
jrochkind: thanks for URL to overridden version!

lib/minimagick/tool.rb
```diff
@@ -0,0 +1,16 @@
...
+  prepend(Module.new do
+    def command
+      if quiet_arg
+        [*executable, *(['-quiet'] + args)]
+      else
+        super
+      end
+    end
+  end)
+end
```
jrochkind: I'm not sure if this will succesfully override, or end up replacing the entire `MiniMagick::Tool` class with this new implementation. Maybe we should call it "tool_override.rb", and just explicitly require it, maybe globally in config/application.rb or something?  I am not sure, thoughts?

jrochkind: or maybe just `lib/minimagick_path.rb`, explicitly required somewhere, to avoid confusion with our automatic `_override` pattern.

hackmastera: You mean you think that the location I put it in will possibly cause it to be loaded when we didn't intend? that makes sense. I like `lib/minimagick_patch`.

jrochkind: I mean, if minimagick itself has a source file at it's own `./lib/minimagick/tool.rb`, then when something tries to load `Minimagick::Tool` (or `require "minimagick/tool"`), it may get out file, and _not_ the original. Our file will end up the sole implementing source for `Minimagick::Tool`, it won't be patching the original, it'll be the whole thing, cause it'll shadow the original from ever being loaded at all.

jrochkind: I think the safe way is

a) make our file at a path that isn't going ot conflict with any "real" file

b) make sure our file refers to Minimagick::Tool first in some way other than just `class MiniMagick::Tool`, some way that will _raise_ if there _is_ no Minimagick::Tool already previously, rather than just defining it to be only what's in this file.

hackmastera:
@jrochkind I updated the overrides to use class_eval, changed the path of the imagemagick one; can you take another look?

jrochkind:
Looks good. Doing it this way with the minimagick patch, I would have thought you would have needed to `require 'minimagick_patch'` somewhere, to ensure it is loaded and applies the patch. Is it working without that though?  

Actually the fact that we have lib/ auto-loaded by Rails makes this all more confusing too, since I believe you aren't supposed to use `require` with auto-loaded files. Gah, I'm not really sure how to do this best.

hackmastera:
I required it in `chf_tasks`, where it used to be inline.

maybe we should explicitly load files in `lib` instead of autoloading them? Strange; it didn't seem to be autoloading. I did write a temporary test and I had to require the file before it would take effect.

jrochkind:
Ah, I see the require now, sorry I missed it!  We could also just require in `config/application.rb` and have it everywhere. But the confusion over `require` to things that are rails autoloaded may suggest leave it how it is for now. 

I can explain why it wasn't being autoloaded. Here's how Rails autoloading works. When you ask for a constant which does not yet exist, Rails will look in it's auto-load paths for a file whose name (and path) matches that constant's name (and namespace) when converted to underscore format. 

So you'd have to ask for a constant called MiniMagickPatch to trigger rails autoloading of this file -- and then Rails would just complain that it expected `minimagick_patch.rb` to define a class called `MiniMagickPatch`!

Autoloading has been the source of much confusion and problems over Rails history, it's really pretty hacky. It's also pretty convenient. 

If we had to do it over again, I think it might be good to not have 'lib' autoloaded. I tend to think of `app` as the place that's autoloaded (so you have to be really careful sometimes), and `lib` as the place that just works like ordinary ruby.   But this is complicated by the fact that much of the stack puts things in lib that I would have put in app instead of lib. And at this point, changing it might break a lot of things in our code expecting they can use things in `lib` without `requiring`, that we probably don't want to spend time fixing, at least right now. 

So for now I think it's prob good how it is!

jrochkind:
And now I remember why I just inlined this in the file to begin with, I got confused trying to think through all this stuff, so just did the simple thing that would work, knowing it'd probably have to be corrected later if we needed the patch in some other place too, heh. I think your solution is a fine improvement for now!

hackmastera:
thanks for the explanation. it all makes sense. something i'd like to clean up another time.

Issue #617, opened 2017-06-15
install riiif


hackmastera:
for reference: https://github.com/RepoCamp/ahc/wiki/IIIF

note we aren't using manifests.

TODOs
- [x] configure caching
- [x] write cleanup jobs
- [x] use indexed original_file id from presenter
- [x] index a fileset's original file id so we don't have to hit fedora
- [x] add configuration to enable us to give client the url of another box
- [x] routing: ideally we don't want riiif routes to be routable on 'sufia' instance, and we _only_ want riiif routes to be routable on 'riiif' instance. So a config variable set by ansible somewhere. But note we still need the routes to be generatable, at least the riiif routes. Needs some experimentation. (jrochkind has a proof of concept, he can do this one)
- [x] solr original_file_id, if not present you shouldn't get a 500. Get a placeholder image or blank image? Resort to the Fedora lookup to find it?  Either way log an error/warning?
- [x] `XMLHttpRequest cannot load http://34.207.123.110/image-service/mp48sd07w%2Ffiles%2F41908f2d-eecf-4720-9bfd-8072fca09477/info.json. No 'Access-Control-Allow-Origin' header is present on the requested resource. Origin 'https://staging.digital.chemheritage.org' is therefore not allowed access. The response had HTTP status code 500.`
- [x] #665 
- [x] https://github.com/chemheritage/chf-sufia/issues/677 fix resque config
- [x] config feature switch to tell show page to use 'jpeg' derivative instead of riiif #686

hackmastera:
i lthink the 2-layer tiffs are going to trip us up again. I ran the command that riiif executes and got 2 files:
```
        test-0.jpg
        test-1.jpg
```

hackmastera:
for those following along on email, the 2-layer tiffs don't appear to be a problem during testing.

hackmastera:
yeah, spoke too soon. looks like it's a problem once you get it into seadragon.

jrochkind:
Did we have a solution for this that just told ImageMagick to collapse into one layer before doing anything else?  There might be an easy way to configure/override riiif to do that, and it sounds like a good PR upstream too.

hackmastera:
yes i was messing around with exactly that this afternoon. not quite there yet.

jrochkind:
What makes the "original_file_id" "original"?  Ah, the "original file".  Cause you need to give a file ID not a filesetid to riiif?  

[...]

PR #634, opened 2017-06-22
[WIP] Riiif

hackmastera:
see #617 

Here for discussion on commits so far, as desired.


app/assets/javascripts/chf_image_viewer.js
```diff
@@ -124,10 +117,11 @@ ChfImageViewer.prototype.selectThumb = function(thumbElement) {
   var shouldShowInfo = thumbElement.getAttribute('data-member-should-show-info') == "true";
   var title = thumbElement.getAttribute('data-title');
   var linkUrl   = thumbElement.getAttribute('data-member-show-url');
+  var titleSource = thumbElement.getAttribute('data-title-source');
 
   $('.viewer-image').addClass('viewer-image-loading');
 
-  this.viewer.open(this.id2TileSource(id));
+  this.viewer.open(titleSource);
```
jrochkind: what's "title" about this?  A typo for `tile-source`?

app/helpers/riiif_helper.rb
```diff
@@ -0,0 +1,12 @@
+module RiiifHelper
+  # use relative url unless we've defind a riiif server in config/environments/*.rb
+  def riiif_image_url (file_set_id)
+    file_id = FileSet.find(file_set_id).original_file.id
+    path = riiif.info_path(file_id, locale: nil)
+    begin
+      return URI.join(Rails.configuration.riiif_server, path).to_s
+    rescue NoMethodError
+      return path
```
jrochkind: what circumstances is this rescuing from, what method ends up missing?  If you don't have `Rails.configuration.riiif_server`?  I think a `Rails.configuration.respond_to?(:riiif_server)` is better than a rescue.

app/assets/javascripts/chf_image_viewer.js
hackmastera: i think my wires were crossed a bit. 'tile source' makes so much more sense; I kept wondering why it was called 'title source'.

app/helpers/riiif_helper.rb
hackmastera: sounds good.

hackmastera:
@jrochkind thanks for the review so far; i believe I've addressed all your points up to now.

jrochkind:
Is this ready for merge, as far as riiif functionality goes?

Before we actually merge to master, we need to deal with purging cached files -- whether it's going to be on the app server or a riiif server with more expansive disk space will effect purge strategies. Either way, could be a cronjob deployed by cap, or something in ansible. But we shouldn't deploy to prod until we have _something_ there, so shouldn't merge to master either, I think?

But we could play with this on staging. And I might work on top of it to do other things I was on waiting for riiif. eg #635 #636

hackmastera:
This PR is against master, but you could work against this branch if you like.

Outstanding TODOs are on https://github.com/chemheritage/chf-sufia/issues/617

All the commits here have been reviewed. would you like me to submit the next one as a PR against this branch?

hackmastera:
But yes, riiif stuff works on this branch you could do use this as a starting point for those other issues you mentioned.

jrochkind:
Ah, I see, so you don't plan to merge this to master until all the checkboxes on #617 are marked?

I don't have clarity myself on whether those should be PR's against this or master, what was your original plan or what do you think would be most convenient?

hackmastera:
@jrochkind right, i wasn't going to merge to master until that ticket is done. Just like you said above: "Before we actually merge to master, we need to deal with purging cached files -- whether it's going to be on the app server or a riiif server with more expansive disk space will effect purge strategies"...

i'm working on the indexing piece right now; nearly done. the caching stuff will be all that's left after that.

I think it will be easier for you to review if I PR against this branch, since you've already reviewed what's here.

config/initializers/riiif.rb
```diff
@@ -0,0 +1,33 @@
...
+  fs_id = id.sub(/\A([^\/]*)\/.*/, '\1')
+  resp = ActiveFedora::SolrService.get("id:#{fs_id}")
+  doc = resp['response']['docs'].first
+  raise "Unable to find solr document with id:#{fs_id}" unless doc
+  { height: doc['height_is'], width: doc['width_is'] }
+end
+
+def logger
+  Rails.logger
+end
```
jrochkind: Missed this before, but this isn't right. You are effectively defining a logger method at top-level ruby scope. Or somewhere else that probably isn't what we want. 

What was intended here? Did this actually come from a riiif-generated template? grrr.

hackmastera: oops, sorry. I pulled it from hyku as per https://github.com/RepoCamp/ahc/wiki/IIIF#install-a-riiif-config

here's the file I used: https://github.com/samvera-labs/hyku/blob/master/config/initializers/riiif.rb

jrochkind: I dunno, I think hyku is very wrong there, don't understand what it's doing.

PR #637, opened 2017-06-26
Use indexed original_file id from presenters

hackmastera:
so viewer doesn't need to hit fedora.

### DIFF
--- a/app/presenters/curation_concerns/generic_work_show_presenter.rb
+++ b/app/presenters/curation_concerns/generic_work_show_presenter.rb
@@ -78,5 +78,11 @@ module CurationConcerns
         end
       end
     end
+
+    def riiif_file_id
+      return unless member_presenters.present?
+      representative_presenter = member_presenters.select{ |pres| pres.id == representative_id }.pop
+      representative_presenter.riiif_file_id
+    end
   end
 end
--- a/app/helpers/riiif_helper.rb
+++ b/app/helpers/riiif_helper.rb
@@ -1,6 +1,6 @@
 module RiiifHelper
   # use relative url unless we've defind a riiif server in config/environments/*.rb
-  def riiif_image_url (riiif_file_id)
+  def riiif_info_url (riiif_file_id)
     path = riiif.info_path(riiif_file_id, locale: nil)
     if Rails.configuration.respond_to? :riiif_server
       return URI.join(Rails.configuration.riiif_server, path).to_s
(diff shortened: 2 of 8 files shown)


app/presenters/curation_concerns/generic_work_show_presenter.rb
```diff
@@ -78,5 +78,11 @@ def show_thumb_member_presenters
         end
       end
     end
+
+    def riiif_file_id
+      return unless member_presenters.present?
+      representative_presenter = member_presenters.select{ |pres| pres.id == representative_id }.pop
+      representative_presenter.riiif_file_id
+    end
```
jrochkind: I'm not totally sure if there might be a quicker way to get the representative presenter than loading _all_ the member presenters and filtering them... maybe not though. 

I'm not a fan of the contortions we've got to go through to get all this stuff, and to keep track of what we should be sending where when. And the performance implications of loading all this stuff. But this may just be what the stack requires.

hackmastera: do you think it would be better to just create the desired presenter?

jrochkind: Hmm. I guess we gotta think through what's happening. 

If all the member_presenters are already loaded, what's here is fine. Do we think they always/usually will be?

If they are not, and we have enough information to just create the desired presetner, it would _probably_ be a lot more performant to just do that, than to create _all_ the member presenters (which might not be needed in this action).  Since I think creating all of the member presenters may require _multiple_ solr requests, one for each?

Hard to know for sure, because I don't completely understand what's going on at the lower levels with fetches from where when. 

What do you think?

jrochkind: If you want one thing, I think `find` is clearer (and slightly more efficient) than `select... pop`.

jrochkind:
I would like to merge this into riiif so I can work on top of it though.

hackmastera:
great feel free to merge and I will look at making that change as another commit.

app/presenters/curation_concerns/generic_work_show_presenter.rb
hackmastera: looks like there's already a way to grab this, which I will use! https://github.com/samvera/curation_concerns/blob/master/app/presenters/curation_concerns/work_show_presenter.rb#L47

PR #655, opened 2017-06-27
riiif-provided srcset images on show page

### LINKED ISSUE #636: proper image sizes in show/viewer
The show page and viewer thumbs are currently just using the derivatives that already existed. 

Once we have riiif, we can make them use just the right size for where they are on page. Ideally using srcset in some way to tell the browser different sizes for different screen resolutions and/or responsive page sizes.

jrochkind:
Closes #636. 

Just did show page, not viewer. On viewer, using the derivative 'thumb' (which I think we need anyway), is about 2x "retina" size already, and relatively small files already, good enough for now I think.

### DIFF
--- a/app/assets/stylesheets/local/show.scss
+++ b/app/assets/stylesheets/local/show.scss
@@ -93,7 +93,7 @@
       width: 100%;
       // keep crazy long aspect ratio images from being
       // so long, although there will be some distortion.
-      max-height: 80vw;
+      max-height: 80vh;
     }
 
     .show-page-member-title {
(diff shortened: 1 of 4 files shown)


jrochkind:
Cause riiif is kinda slow, this does lead to noticeably slower image loading at least in my dev copy -- at least until the particular derivatives are cached. 

There are a variety of possible things we can do if this is a problem in production.

* speed up riiif via a number of things
* pre-cache the images by pinging riiif on image upload (depending on cache size), also could use a CDN
* generate these in advance with plain old derivatives instead of riiif
* ameliorate slow loading image problems with spinner and/or pre-calculated height so the page doens't jump around when it loads (we know aspect ratio from characterization, we can calculate what the height will be, although adding this into our CSS might be tricky when image width changes responsively)

hackmastera:
👍  regarding the code

To what extent do you think we need to address your additional discussion points before this is merged?

jrochkind:
I don't think we need to address before merge, I propose we get it in, and then see how much of a problem the riiif performance is in what ways, once we have riiif fully set up.

PR #656, opened 2017-06-28
Add new characterize job that doesn't call derivatives, cleans up aft…

hackmastera:
…er itself, refs #624

### DIFF
--- a/app/jobs/just_characterize_job.rb
+++ b/app/jobs/just_characterize_job.rb
@@ -1,0 +1,38 @@
+# Copy of https://github.com/samvera/curation_concerns/blob/v1.7.7/app/jobs/characterize_job.rb
+#   with call to CreateDerivativesJob removed
+#   and file cleanup added
+class JustCharacterizeJob < ActiveJob::Base
+  include CurationConcerns::Lockable
+
+  queue_as CurationConcerns.config.ingest_queue_name
+
+  # @param [FileSet] file_set
+  # @param [String] file_id identifier for a Hydra::PCDM::File
+  # @param [String, NilClass] filepath the cached file within the CurationConcerns.config.working_path
+  def perform(file_set, file_id, filepath = nil)
+    filename = CurationConcerns::WorkingDirectory.find_or_retrieve(file_id, file_set.id, filepath)
+    raise LoadError, "#{file_set.class.characterization_proxy} was not found" unless file_set.characterization_proxy?
+
+    # Prevent other jobs from trying to modify the FileSet at the same time
+    acquire_lock_for(file_set.id) do
+      Hydra::Works::CharacterizationService.run(file_set.characterization_proxy, filename)
+      Rails.logger.debug "Ran characterization on #{file_set.characterization_proxy.id} (#{file_set.characterization_proxy.mime_type})"
+      file_set.characterization_proxy.save!
+      file_set.update_index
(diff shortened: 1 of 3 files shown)


app/jobs/just_characterize_job.rb
```diff
@@ -0,0 +1,38 @@
...
+    acquire_lock_for(file_set.id) do
+      Hydra::Works::CharacterizationService.run(file_set.characterization_proxy, filename)
+      Rails.logger.debug "Ran characterization on #{file_set.characterization_proxy.id} (#{file_set.characterization_proxy.mime_type})"
+      file_set.characterization_proxy.save!
+      file_set.update_index
+      file_set.parent.in_collections.each(&:update_index) if file_set.parent
+    end
+
+    File.delete(filename)
+    dir = File.split(filename).first
```
jrochkind: Thsi is deleting a file that comes from `find_or_retrieve` -- are we sure this is in fact a temporary copy, and it won't sometimes be accidentally deleting the original?

app/jobs/just_characterize_job.rb
```diff
@@ -0,0 +1,38 @@
...
+    cleanup_directory(dir)
+  end
+
+  def cleanup_directory(dir)
+    return if dir == "tmp"
+    parent = File.split(dir).first
```
jrochkind: why return if it's tmp?  What's going on here?  This stuff is all confusing to me as to what's what and why, maybe some comments with what you've figured out?

app/jobs/just_characterize_job.rb
hackmastera: Great question. It always will be in our use case; I can even delete the parameter that would allow a local original.

Maybe I'll call it Recharacterize instead of JustCharacterize.

jrochkind: "It always will be in our use case" -- even in dev?  Yeah, I'm worried our use case could change in the future and then this could trip us up. But you think you've changed it in a way that's unlikely?  I admit I don't completely follow, but I trust ya, just bringing up for consideration!

app/jobs/just_characterize_job.rb
jrochkind: Is this hard-coding to assume these files will always be in a directory called `tmp`?  (It doesn't actually insist they are in `/tmp`, it can be any directory called `tmp`, including `/tmp` or local `./tmp`).  Is this a safe assumption, might we want to move these at some point?  Is there an actual existing config (from sufia) you can take this parent dir name from, rather than assuming /tmp?   (and then check full-path, not just dir name)?

hackmastera: Fixed. Thank you!

app/jobs/just_characterize_job.rb
hackmastera: Yep, totally good / fair question. This job will always pull a file out of fedora into a 'working' space, characterize it, and then delete the temporary file. Even in dev there is not a cache of fedora originals sitting around for this purpose. And now that I've deleted the parameter that allowed for a local file, we couldn't pass one in even if we had one.

The only possible conflict would be if some other job had pulled the file into the working space and was doing something to it when this one cleaned it up. That's a risk that I (uncharcteristicallly!) am willing to take, since we will run this very infrequently if ever again. The only place this job is called is from the 'recharacterize' rake task.

Here, I'll add another comment on the job and the task.

PR #658, opened 2017-06-28
A CHF::Env class that we will access infrastructure/environmental config from.

jrochkind:
This pulls from the config/local_env.yml if it's present. 

BUT it lets you override that with actual ENV -- awfully useful for testing in dev to make
sure it works how you want in different setings. Also useful if we want to deploy
the app to another environment like heroku that actually uses ENV. 

And it also lets you set a default value, including as a lambda using other env
or config, for us to set our default dev values, or calculate values that aren't
yet in our local_env.yml as independent values. 

The actual implementation of CHF::Env is a bit hacky, although not too complicated,
but I think the API/semantics are pretty good and should be pretty stable. 

In addition, this PR then uses the new CHF::Env for:

* public_riiif_url
* setting riiif file cache paths, with the defaults that we want (there is also
  clear riiif file caches rake task)
* turning on routing of riiif/app paths with defaults that will still
  route everything in dev or if otherwise unset, but route only
  the right things if `service_level` is properly set.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,76 +1,85 @@
 require 'resque/server'
 
 Rails.application.routes.draw do
-  # override sufia's about routing to use a static page instead of a content block
-  get 'about', controller: 'static', action: 'about', as: 'about'
+  # RIIIF server should serve ONLY riiif paths; app server anything BUT riiif paths
+  # development server (or any other server you configure cause you want it) can still serve both
 
-  concern :range_searchable, BlacklightRangeLimit::Routes::RangeSearchable.new
-  Hydra::BatchEdit.add_routes(self)
-  mount Qa::Engine => '/authorities'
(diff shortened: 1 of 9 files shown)


config/routes.rb
```diff
@@ -1,76 +1,85 @@
...
+    mount Hydra::RoleManagement::Engine => '/'
 
-  Hydra::BatchEdit.add_routes(self)
-  # Sufia should be mounted before curation concerns to give priority to its routes
-  mount Sufia::Engine => '/'
-  mount CurationConcerns::Engine, at: '/'
+    get '/focus/:id', to: 'synthetic_category#show', as: :synthetic_category
+
+
+    Hydra::BatchEdit.add_routes(self)
```
hackmastera: unrelated to this PR, but this line is in the routes twice. I guess we should remove the later one.

PR #662, opened 2017-06-29
Use Addressable::URI.join to allow for schema-agnostic urls

hackmastera:


jrochkind:
Does this code work if `CHF::Env.lookup(:public_riiif_url)` is a bare host, or a schema-relative `//host`?  

Would it make sense to have tests demonstrating?

hackmastera:
works with `//`; didn't try bare host as per slack discussion.

I can work on a test.

jrochkind:
up to you, if you don’t think the benefit is worth the pain/time, your call.

jrochkind:
FWIW, tested just to see what would happen. Addressable::URI.join ignores the first arg if it's just a bare hostname/IP, I dunno why that's what it chooses!

~~~
[7] pry(main)> Addressable::URI.join("123.123.123.123", "/foo").to_s
=> "/foo"
[8] pry(main)> Addressable::URI.join("example.com", "/foo").to_s
=> "/foo"
~~~

jrochkind:
Maybe we actually want to pull this out into another method to make_riiif_url_with_host or something, that can be called in both places, and can be written as defensively as we like (or checking args and immediately raising if the args aren't legal, or wahtever).

hackmastera:
interesting!

this test is almost done.

hackmastera:
So if the `//` is missing it will just return the relative path. I guess that wouldn't work since we're turning off that routing. You think it would be better to raise? or if we want to validate the config, maybe it would be better to do it in the config class? not sure where the right seam is.

jrochkind:
I guess I could add a validation feature to the config class. I am not sure if that makes sense, maybe?

Not really sure either!

I like that we turned routing off, so if it's not doing what we expect, it just _doesn't work_ instead of routing from the local app and not using the riiif server at all without us noticing!

hackmastera:
i'll make it raise; working on it now; won't take long.

jrochkind:
changed argument error message to be more accurate and helpful. merge at will!

PR #666, opened 2017-06-30
Empty child fix

hackmastera:
I tried this 2 ways; this way seemed a little messy but it works.

The other (see https://github.com/chemheritage/chf-sufia/commit/b8f199e7551a6c4b5ef15bfd8b724cd0e55497e1 on branch empty-child-fix-2) was an attempt to handle it as much as possible in the helper, but it resulted in the viewer failing to show the default thumb, plus there was still a new conditional needed in the download link logic.


hackmastera:
self-merged, but ready for review after the holiday and can make changes if needed

PR #670, opened 2017-06-30
Take the max of integers, not strings (doesn't interpolate i guess)

hackmastera:


hackmastera:
I will self-merge once travis approves. fee free to review after holiday.

PR #678, opened 2017-07-10
Authorize for riiif image access, refs #617

hackmastera:
as per https://github.com/RepoCamp/ahc/wiki/IIIF#adding-authorization


hackmastera:
actually, they have a test, too; let me steal that over as well.

hackmastera:
test now included

PR #679, opened 2017-07-11
Rake task to get riiif to preload originals

jrochkind:
eg, on app server:

    RAILS_ENV=production INTERNAL_RIIIF_URL=http://172.31.63.136 bundle exec rake chf:riiif:preload_originals


jrochkind:
oops, not supposed to be to master.

PR #680, opened 2017-07-11
lazysizes-aspectratio on all show page images and viewer thumbs

### LINKED ISSUE #669: show page - primary image container should be pre-sized

jrochkind:
Takes up proper space on page even before image has loaded, since we tell it expected
aspect ratio, gotten from recorded characterization.

works whether the images is being lazy loaded or normal loaded, nice!
lazysizes-aspectratio is a nice script.

Closes #669

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -12,6 +12,11 @@
 
 //= require jquery
 //= require jquery_ujs
+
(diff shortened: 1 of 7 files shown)


hackmastera:
nice, very clean, uses a dependency we already introduced earlier. Thanks!

PR #683, opened 2017-07-11
Sort out google analytics semantics, pass desired data; fixes #668

### LINKED ISSUE #668: Ensure GA download data is in a usable form
Once we've deployed the new events to prod, @MDiMeo should peruse GA and ensure the data is useful.

hackmastera:
Variable names in the generalized javascript have been changed to match google analytics. The generalized js now requires a class, and looks up the element's hierarchy to find it if it's not on the target element. This allows the container div to hold the ga data, but still use the generalized javascript.

### DIFF
--- a/app/assets/javascripts/chf_ga_events.js
+++ b/app/assets/javascripts/chf_ga_events.js
@@ -5,9 +5,10 @@
 //
 // https://developers.google.com/analytics/devguides/collection/gajs/eventTrackerGuide
 
-$(document).on('click', '*[data-analytics-event]', function(e) {
+$(document).on('click', '*[data-analytics-category]', function(e) {
   _gaq.push(['_trackEvent',
-             e.target.getAttribute("data-analytics-event"),
+             e.target.getAttribute("data-analytics-category"),
+             e.target.getAttribute("data-analytics-action"),
              e.target.getAttribute("data-analytics-label"),
              e.target.getAttribute("data-analytics-value")
             ]);
--- a/app/assets/javascripts/chf_image_viewer.js
+++ b/app/assets/javascripts/chf_image_viewer.js
@@ -451,13 +451,6 @@ jQuery(document).ready(function($) {
       event.preventDefault();
       var id = this.getAttribute('data-member-id');
       chf_image_viewer().show(id);
-
-      // GA
-      _gaq.push(['_trackEvent',
-           'ImageViewer',
-           'fileSetId',
-           id
-          ]);
     });
 
     // with keyboard-tab nav to our thumbs, let return/space trigger click as for normal links
--- a/app/helpers/riiif_helper.rb
+++ b/app/helpers/riiif_helper.rb
@@ -38,7 +38,7 @@ module RiiifHelper
   #
   # if use_image_server is false, size_key is ignored and no srcsets are generated,
   # we just use the stock hydra-derivative created image labelled 'jpeg'
-  def member_image_tag(parent_id, member, size_key: nil, lazy: false, use_image_server: CHF::Env.lookup(:use_image_server_on_show_page))
+  def member_image_tag(parent_id:, member:, size_key: nil, lazy: false, use_image_server: CHF::Env.lookup(:use_image_server_on_show_page))
     base_width = size_key == :large ? 525 : 208
 
     args = {
--- a/app/helpers/riiif_helper.rb
+++ b/app/helpers/riiif_helper.rb
@@ -38,7 +38,7 @@ module RiiifHelper
   #
   # if use_image_server is false, size_key is ignored and no srcsets are generated,
   # we just use the stock hydra-derivative created image labelled 'jpeg'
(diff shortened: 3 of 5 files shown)


app/assets/javascripts/chf_ga_events.js
```diff
@@ -2,13 +2,15 @@
 // https://github.com/samvera/hyrax/blob/1e504c200fd9c39120f514ac33cd42cd843de9fa/app/assets/javascripts/hyrax/ga_events.js
 //
 // We add our own more flexible.
+// Can be used by a container or directly by a target; just requires the 'ga-event-data' class on the element holding the data.
 //
 // https://developers.google.com/analytics/devguides/collection/gajs/eventTrackerGuide
 
-$(document).on('click', '*[data-analytics-event]', function(e) {
+$(document).on('click', '.ga-event-data', function(e) {
   _gaq.push(['_trackEvent',
```
jrochkind: Where possible/convenient (and it isn't always), I like using [data] attributes for javascript hooks, instead classes -- classes are for CSS styling, data attributes are for JS hooks, which are different things. 

Did you make this change intentionally, or can we go back to a data hook?  I liked naming the data hook generically after 'analytics' not 'ga' specifically, hypothetically it could be used for alternative analytic services too, but I feel less strongly about that if you think naming it ga is less confusing.

app/assets/javascripts/chf_ga_events.js
```diff
@@ -2,13 +2,15 @@
...
-             e.target.getAttribute("data-analytics-event"),
-             e.target.getAttribute("data-analytics-label"),
-             e.target.getAttribute("data-analytics-value")
+             e.target.closest('.ga-event-data').getAttribute("data-analytics-category"),
+             e.target.closest('.ga-event-data').getAttribute("data-analytics-action"),
+             e.target.closest('.ga-event-data').getAttribute("data-analytics-label"),
+             e.target.closest('.ga-event-data').getAttribute("data-analytics-value")
             ]);
```
jrochkind: thanks!

app/assets/javascripts/chf_image_viewer.js
```diff
@@ -438,13 +438,6 @@ jQuery(document).ready(function($) {
...
       var id = this.getAttribute('data-member-id');
       chf_image_viewer().show(id);
-
-      // GA
-      _gaq.push(['_trackEvent',
-           'ImageViewer',
-           'fileSetId',
-           id
-          ]);
     });
```
jrochkind: You replaced this with the data attributes on the link?  If that works properly, that is better! I was worried it wouldn't because of the fancy JS stuff being done in to trigger this click, instead of using an ordinary hyperlink, so instead of doing the right thing and testing and getting it right, I was just lazy and did something I knew would work.

app/assets/javascripts/chf_ga_events.js
hackmastera: Yes, I tried to make it work that way but because we were putting the data attributes on the containing div instead of on the image / target elements it wasn't picking them up. I tried a couple of things to preserve use of data attributes this way, but I couldn't get it to work. The only alternatives I saw were:
* replicate the data elements into 3 branches of a conditional. (yuck)
* keep the separate call to `_gaq.push` in chf_image_viewer.js. I preferred to move this out and have all the ga events handled by this js file, having earlier missed the separate call entirely because it didn't use this convention.

app/assets/javascripts/chf_image_viewer.js
hackmastera: Yep, tested and working as long as we are okay with the requirement for a class instead of being able to use just data attributes.

app/assets/javascripts/chf_ga_events.js
hackmastera: if you're okay with this choice, I can change the name of the class if you like. `.log-analytics-event`?

hackmastera:
@jrochkind updated to use your new helper; class no longer required. take another look?

app/helpers/riiif_helper.rb
```diff
@@ -38,7 +38,7 @@ def riiif_image_srcset_pixel_density(riiif_file_id, base_width, format: 'jpg', q
   #
   # if use_image_server is false, size_key is ignored and no srcsets are generated,
   # we just use the stock hydra-derivative created image labelled 'jpeg'
-  def member_image_tag(member, size_key: nil, lazy: false, use_image_server: CHF::Env.lookup(:use_image_server_on_show_page))
+  def member_image_tag(parent_id, member, size_key: nil, lazy: false, use_image_server: CHF::Env.lookup(:use_image_server_on_show_page))
     base_width = size_key == :large ? 525 : 208
```
jrochkind: gah, I guess we have to pass parent_id in too now for the analytics. :(  Should we make them all keyword args to avoid confusion as to which is which?

jrochkind: wait, are you actually using the parent_id you pass in at all? Did you mean to be?

hackmastera: sure, see new commit.

PR #687, opened 2017-07-12
App will use images on show page from riiif or hydra-derivative 'jpeg' depending on config

### LINKED ISSUE #686: config feature switch to tell show page to use 'jpeg' derivative instead of riiif
backup plan, and even once deployed if riiif is not working out lets us switch back and forth. 

The existing 'jpeg' hydra-derivative isn't quite the right resolution -- it's too small for the big representative image, and too big for the smaller thumbs. And lacks alternatives for srcset resolutions. 

But it works, so if the riiiif is not, it can be used as a fallback that keeps our app functional!

jrochkind:
To accomplish, also refactored _show_page_image image tag generation to a
unified member_image_tag helper. Which will make it easier to do the GA
stuff in #668

Currently defaults to using riiif. If we don't want to use riiif, we could
change the default, or just add a key in local_env.yml,

    use_image_server_on_show_page: false

You can experiment in dev to see both modes by starting up as:

    USE_IMAGE_SERVER_ON_SHOW_PAGE=false rails server
    USE_IMAGE_SERVER_ON_SHOW_PAGE=true rails server

Closes #686

### DIFF
--- a/app/helpers/riiif_helper.rb
+++ b/app/helpers/riiif_helper.rb
@@ -35,6 +35,9 @@ module RiiifHelper
   # for use on show page. Calculates proper image tag based on lazy or not,
   # use of riiif for images or not, and desired size. Includes proper
   # attributes for triggering viewer, analytics, etc.
+  #
+  # if use_riiif is false, size_key is ignored and no srcsets are generated,
+  # we just use the stock hydra-derivative created image labelled 'jpeg'
   def member_image_tag(member, size_key: nil, lazy: false)
     base_width = size_key == :large ? 525 : 208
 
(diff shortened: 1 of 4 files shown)


jrochkind:
When configured not to use riiif, it uses the existing 'jpeg'-keyed hydra-derivative derivative. Not quite the right resolution (too small for representative, too big for thumbs), no srcset, but it works.

app/helpers/riiif_helper.rb
```diff
@@ -31,6 +31,52 @@ def riiif_image_srcset_pixel_density(riiif_file_id, base_width, format: 'jpg', q
...
+  # we just use the stock hydra-derivative created image labelled 'jpeg'
+  def member_image_tag(member, size_key: nil, lazy: false, use_image_server: CHF::Env.lookup(:use_image_server_on_show_page))
+    base_width = size_key == :large ? 525 : 208
+
+    args = {
+      class: ["show-page-image-image"],
+      alt: "",
+      data: {
+        trigger: "chf_image_viewer",
+        member_id: member.id,
```
hackmastera: Just noticed this, though it's already merged. One case used `member.representative_id`; no longer needed?

jrochkind: uh... i'm not sure which is right. I think thsi is right now and the other one was wrong, though. Should have collapsed those branches before, too many opportunities for inconsistency and bugs!

jrochkind: it's def possible we'll still find bugs relating to this code. we'll fix em when we find em, I dunno.

PR #693, opened 2017-07-13
fix click on child work to launch viewer

jrochkind:
@hackmastera  spotted the bug from code alone at:
https://github.com/chemheritage/chf-sufia/pull/687#discussion_r127051821
go you.


hackmastera:
Tried this on my dev box and it looks good! Thanks!

jrochkind:
I am pleased that the code was resilient enough that the bug was just "we don't know what image you're asking for, we give you the first one", and not a 500 or JS console error breaking JS code.

hackmastera:
Yeah that is really super good!! Excellent work.

PR #696, opened 2017-07-13
Add original_file_id indexing to master.

jrochkind:
Copy of what is in riiif branch. Will have to resolve on merging master/riiif.
We want this here so master can be indexing what riiif will need, before
we deploy riiif


hackmastera:
great, thanks!

PR #698, opened 2017-07-14
Consolidate solr queries for members, reduces load times for large works

hackmastera:


app/presenters/curation_concerns/member_presenter_factory_override.rb
```diff
@@ -6,3 +6,24 @@
...
+  # Override to get all member solr docs in one solr query, rather than getting them one at a time in a presenter factory
+  def member_presenters(ids = ordered_ids, presenter_class = composite_presenter_class)
+    @member_presenters ||=
+      begin
+        ordered_docs.select { |x| ids.include?(x.id) }.map do |doc|
+          presenter_class.new(doc, *presenter_factory_arguments)
+        end
+      end
+  end
+
```
jrochkind: I am not a fan of the member_presenters override, I expect fragility, but I agree this is the right approach if the base code wasn't doing it all in one solr query and this makes it do so. :(

jrochkind: It's kind of ridiculous this isn't the default behavior.

hackmastera: Yeah, a contribution back to hyrax was requested while discussing on slack

Issue #700, opened 2017-07-17
improve show page load time performance on large works


hackmastera:
#698 was a start but we need to do more

hackmastera:
I'd like to also consider what we can push upstream, even if it's only tickets that reference our fixes.

jrochkind:
For sample data: Ramelli in production has 694 members, 196 of which are child works, the others are filesets.

jrochkind:
I made a test record with 50 child works and 0 file sets. 

On riiif branch, it makes 164 requests to solr to render the show page. 

On master branch, only 11. 

So that's probably some of it, the patch to prevent n+1 solr queries doesn't work for child works somehow -- but only in our riiif branch. Maybe riiif branch is asking for something that triggers a solr fetch, that master branch is not. Investigating further.

jrochkind:
having trouble getting width and height into the index. Not sure if it's an artifact of dev:data, or if I'm looking in the wrong place. 

Hmm, none of my filesets have "width" or "height" set on them in fedora. Weird. Are they supposed to?

jrochkind:
yep, just an artifact of my setup. ignore, fixing it.

jrochkind:
I think we did as much of this as we are going to do for now, and can close? Yes @hackmastera ?

#723, #724, #720, #704, #705

PR #704, opened 2017-07-18
override Solrizer to cache, speed up

jrochkind:
Solrizer.solr_name is called a lot on our show page.
Profiling revealed it was somehow a bottleneck. This helped.


hackmastera:
refs #700

PR #705, opened 2017-07-18
look up rights metadata for rights statements faster

jrochkind:
We were using , from QA gem.

This ended up re-loading YAML from disk on every call. Either the QA code needs caching
it doesn't have, or we're using QA wrong, or both.  Rather than figure it out,
we write our own simple wrapper to get the info from the same .yml file,
and make sure it's cached globally. Nothing fancy about this no reason
not to write ourselves.

### DIFF
--- a/app/presenters/curation_concerns/generic_work_show_presenter.rb
+++ b/app/presenters/curation_concerns/generic_work_show_presenter.rb
@@ -22,8 +22,9 @@ module CurationConcerns
 
     def rights_icon(identifier = rights.first)
       # schema allows multiple rights, we only use one.
-      # we have added our own metadata to the liceneses.yml that the CC
-      # class doens't really have accessors for, we'll kinda hack it.
+      # we have added our own metadata to the licenses.yml used by
+      # CC/QA, but the CC/QA code makes it difficult to access
+      # efficiently, so we use our own CHF::RightsTerms
       case CHF::RightsTerms.category_for(identifier)
         when "in_copyright"
           "rightsstatements-InC.Icon-Only.dark.svg"
--- a/app/presenters/curation_concerns/generic_work_show_presenter.rb
+++ b/app/presenters/curation_concerns/generic_work_show_presenter.rb
@@ -24,15 +24,15 @@ module CurationConcerns
       # schema allows multiple rights, we only use one.
       # we have added our own metadata to the liceneses.yml that the CC
       # class doens't really have accessors for, we'll kinda hack it.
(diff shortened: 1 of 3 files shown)


jrochkind:
Before PR, we were looking up stuff from .yml file via CC/QA with calls like this:

    CurationConcerns::LicenseService.new.authority.find(identifier)

Profiling suggested that there might have been a lot of time spent in `YAML.load`. Was this reloading YAML from disk all the time instead of using a cached copy?

Looking at stack traces and QA code (we are currently using qa 0.11.1), I first noticed this:

https://github.com/samvera/questioning_authority/blob/v0.11.1/lib/qa/authorities/local/file_based_authority.rb#L28

Indeed there is no caching going on, every time you ask this authority to [find](https://github.com/samvera/questioning_authority/blob/v0.11.1/lib/qa/authorities/local/file_based_authority.rb#L21) or [search](https://github.com/samvera/questioning_authority/blob/v0.11.1/lib/qa/authorities/local/file_based_authority.rb#L9) it calls `#terms`... which reads and parses the YAML to disk. 

At first I considered adding caching here. But adding caching to FileBasedAuthority only works if you are re-using a file-based authority.  The API we are using started with `CurationConcerns::LicenseService.new.authority.find(identifier)`, which created a new FileBasedAuthority object each time. Verified by putting some logging/byebug in there, indeed FileBasedAuthority#initialize was called many many times per our multi-member `show` page. 

It's possible there is some other API in CC or QA we could use that would re-use a FileBasedAuthority object instead of creating a new one each time. This stuff isn't documented very well, I'm not certain. Even if it were, FileBasedAuthority is _still_ not caching. 

I decided QA was just not intended for this use case (or any use case but auto-completing an edit field via an AJAX request), doesn't neccesarily have the API or semantics to support it, and rather than try to figure it out and/or hack it to do so -- reading a YAML file is not a complicated thing, it made sense just to spend the 45 minutes writing the ~40 straightforward lines of code to just read the YAML file and cache it.

jrochkind:
On the mutant riiif branch I was working on, this took my 200-member demo record from around 2.1s to around 1.8s.

hackmastera:
refs #700

app/presenters/curation_concerns/generic_work_show_presenter.rb
```diff
@@ -24,15 +24,15 @@ def rights_icon(identifier = rights.first)
       # schema allows multiple rights, we only use one.
       # we have added our own metadata to the liceneses.yml that the CC
       # class doens't really have accessors for, we'll kinda hack it.
```
hackmastera: Is this comment out of date now?

PR #722, opened 2017-07-19
Environment configuration to use in ruby profiling

hackmastera:


Gemfile
```diff
@@ -107,3 +107,7 @@ end
 group :development, :test do
   gem 'fcrepo_wrapper'
 end
+
+group :profile do
+  gem 'ruby-prof'
```
jrochkind: can we include ruby-prof in dev and test too? Any downsides?  I guess we could do it with `require: false` just to make _sure_ it's having no effect at all in those envs (would have to manually `require` it if you wanted to use it). Don't know if tha'ts neccesary, it ought to have no effect if not being used, and it's not in production.

config/environments/profile.rb
```diff
@@ -0,0 +1,41 @@
+# an environment configured for performance profiling.
```
jrochkind: i'm curious where you got this from. It does not exactly match what we have as environments/production.rb, any reason not to just copy (or even load) environments/production.rb, so we know we're profiling under our actual production config?

jrochkind: Here's the line from our environments/production.rb which I think is the one you were asking about, and is not here. 
https://github.com/chemheritage/chf-sufia/blob/master/config/environments/production.rb#L27

hackmastera: There were a few things I wanted from dev. I can copy prod instead and put exceptions clearly labeled at the top.

Gemfile
hackmastera: sure

config/environments/profile.rb
hackmastera: now that i have it set up i'm not going to worry about showing errors, etc. new version is  just prod but with inline job queue.

hackmastera: Hm. It looks like serve_static_files is [deprecated](http://edgeguides.rubyonrails.org/5_0_release_notes.html#railties-deprecations), and we would use [public_file_server.enabled](http://guides.rubyonrails.org/configuring.html#rails-general-configuration) instead. Default is true except in production environment. I think we need it to stay true since we aren't using an independent web server in this context. so I won't do anything special with that one.

hackmastera:
Self-merging as initially suggested so I can pull into riiif branch.

PR #724, opened 2017-07-20
Move sub-partials called by show_page_image to helper methods instead

jrochkind:
This actually seemed to have a pretty huge impact on performance for
many-membered works. @hackmastera curious if you can confirm/replicate this finding. 

I think it ends up not actually THAT much less legible by the end.

With these additional perf PR's to riiif branch, I think there's a chance riiif branch may actually become FASTER than master branch at present!


app/helpers/member_helper.rb
```diff
@@ -0,0 +1,121 @@
...
+    list_elements << '<li class="dropdown-header">Download this image</li>'.html_safe
+
+    list_elements << dropdown_menuitem(
+                      link_to("Original Image", ( member ? main_app.download_path(member.representative_id) : "#" ),
+                        target: "_new",
+                        id: "file_download",
+                        data: {
+                          content_hook: "dl-original-link",
+                          # These action/labels are what sufia/hyrax uses, although
+                          # a bit weird. https://github.com/samvera/hyrax/blob/1e504c200fd9c39120f514ac33cd42cd843de9fa/app/assets/javascripts/hyrax/ga_events.js
```
hackmastera: oops, old comment; we should remove since we've changed them. can be a follow-on PR.

hackmastera:
I resolved conflicts; left the CSS for `rights-statement-inline` outside of `.work-show`; I *think* that is what was intended as an end state.

PR #728, opened 2017-07-21
Use the asset pipeline in profile env since no asset compilation has …

hackmastera:
…been done


hackmastera:
self-merging; fixes profile environment bug introduced in #722

PR #730, opened 2017-07-24
Index representative original_file_id, width, and height

jrochkind:
For use on show page. This branch doesn't use them on show page yet, but we need to index
them here on master, before we merge riiif in


jrochkind:
Master deployed to production, bulk reindex on production kicked off in screen.

PR #735, opened 2017-07-25
fix work indexer for representatives, when representative is a work

jrochkind:


hackmastera:
I can't poke any holes in it; let's give it a shot

PR #736, opened 2017-07-26
improve work indexing of representative data

jrochkind:
With tests for edge cases I could think of.
Ended up simpler code doing some TDD-ish dev, hooray.

Tricky taking care of reindexing a work when it has a child work as a representative, and that child work's representative has changed. Have to put some expensive solr lookups in indexing. 

Tests are pretty slow, maybe another 30s to these new tests. Best I could come up with so far. Wary of mocking too much stuff that I'm not super confident won't change in a future version, causing tests to false pass.

### DIFF
--- a/spec/indexers/generic_work_indexer_spec.rb
+++ b/spec/indexers/generic_work_indexer_spec.rb
@@ -76,6 +76,7 @@ RSpec.describe GenericWorkIndexer do
       end
     end
 
+    # This is a mess and very slow, better way to test?
     describe "work with representative child work" do
       let(:child_work) do
         FactoryGirl.create(:work, :real_public_image) do |w|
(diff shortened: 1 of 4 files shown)


spec/indexers/generic_work_indexer_spec.rb
```diff
@@ -94,6 +95,33 @@
...
+          work.save!
+        end
+
+        it "updates parent work in index" do
+          child_work.ordered_members << new_file_set
+          IngestFileJob.perform_now(new_file_set, (Rails.root + "spec/fixtures/sample.jpg").to_s, nil)
+          child_work.representative = new_file_set
+
+          indexed_parent = SolrDocument.find(work.id)
+          expect(indexed_parent["representative_original_file_id_tesim"]).not_to include(new_file_set.original_file.id)
```
hackmastera: do we want contain_exactly instead of include?

jrochkind: `not_to contain_exactly` does not mean the same thing as `not_to include`. Like it could include those things, but not _exactly_, and still pass. 

I guess I could say I expect it to eq `[]` if that's what I expect. I guess I'll look more into what I expect.

jrochkind:
deployed to staging, started `rake chf:reindex` with this on staging.

jrochkind:
Indexing completed on staging, without errors. Completed in 2 hours 50 minutes. 

I realize this is potentially doing a bit of extra work on bulk-reindex, due to the code to trigger an index on 'parent' work using a child work as a representative, when a child work is reindexed.  Some works could end up reindexed multiple times in a bulk reindex. 

Still 2 hours 50 minutes seems _about_ what we expect for a bulk reindex, not sure if it's worth trying to fix.

PR #739, opened 2017-07-31
Generalize `riiif` into `image_service` and add cantaloupe switch

hackmastera:
First pass at generalizing away from the use of `riiif`. Adds switches for using cantaloupe to the path logic in the helper as well as the source cache preload task.

### DIFF
--- a/app/helpers/image_service_helper.rb
+++ b/app/helpers/image_service_helper.rb
@@ -47,9 +47,9 @@ module ImageServiceHelper
   # any responsiveness page layout. Sends somewhat more bytes when needed at some responsive
   # sizes, but way simpler to implement; keep from asking riiiif for even more varying resizes;
   # prob good enough.
-  def riiif_image_srcset_pixel_density(riiif_file_id, base_width, format: 'jpg', quality: 'default')
+  def riiif_image_srcset_pixel_density(file_id, base_width, format: 'jpg', quality: 'default')
     [1, BigDecimal.new('1.5'), 2, 3, 4].collect do |multiplier|
-      iiif_image_url(riiif_file_id, format: "jpg", size: "#{base_width * multiplier},") + " #{multiplier}x"
+      iiif_image_url(file_id, format: "jpg", size: "#{base_width * multiplier},") + " #{multiplier}x"
     end.join(", ")
   end
 
@@ -76,15 +76,15 @@ module ImageServiceHelper
       }
     }
 
-    src_args = if member.riiif_file_id.nil?
+    src_args = if member.representative_file_id.nil?
       # if there's no image, show the default thumbnail (it gets indexed)
       {
         src:  member.thumbnail_path
       }
     elsif use_image_server
       {
-        src: iiif_image_url(member.riiif_file_id, format: "jpg", size: "#{base_width},"),
-        srcset: riiif_image_srcset_pixel_density(member.riiif_file_id, base_width)
+        src: iiif_image_url(member.representative_file_id, format: "jpg", size: "#{base_width},"),
+        srcset: riiif_image_srcset_pixel_density(member.representative_file_id, base_width)
       }
     else
       {
--- a/app/helpers/image_service_helper.rb
+++ b/app/helpers/image_service_helper.rb
@@ -1,11 +1,17 @@
-module RiiifHelper
+module ImageServiceHelper
 
   # Returns the IIIF info.json document, suitable as an OpenSeadragon tile source/
   #
   # Returns relative url unless we've defind a riiif server in config/environments/*.rb
-  def riiif_info_url (riiif_file_id)
-    path = riiif.info_path(riiif_file_id, locale: nil)
-    create_riiif_url(path)
+  def iiif_info_url(image_file_id)
+    service = CHF::Env.lookup(:image_server)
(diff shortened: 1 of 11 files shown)


jrochkind:
Can we make this code have fewer custom switches if we include more of the base path in the `local_env.yml`?

I'm not sure if I'm following this right, but would it work if 

    image_server: //cantelope.somewhere/iiif/2/
    # vs
    image_server: //riiif.somewhere/image-server/

And then we adjust the code to add on a relative path to a base path, no longer using the rails helpers for riiif, just doing generic "add iiif api relative path to given base path"?

app/helpers/image_service_helper.rb
```diff
@@ -16,18 +22,34 @@ def riiif_info_url (riiif_file_id)
...
+    create_iiif_url(path)
+  end
+
+  def cantaloupe_info_path(file_id)
+    "iiif/2/#{CGI.escape(file_id)}/info.json"
+  end
+
+  def cantaloupe_image_path(file_id, size:, format:, quality:)
+    region = 'full'
+    rotation = '0'
```
jrochkind: id make 'region' a keyword arg that defaults to `full` instead of hard-coding to full. Maybe rotation too, although we prob have less of a need for it.  I think this is how the riiif helpers basically work, optional args, but you can still supply them. 

Or, I suppose there's an argument for _not_ doing what we just suggested if we don't need it yet.

app/helpers/image_service_helper.rb
```diff
@@ -82,7 +104,7 @@ def member_image_tag(parent_id:, member:, size_key: nil, lazy: false, use_image_
 
   private
 
-  def create_riiif_url(path)
+  def create_iiif_url(path)
     if CHF::Env.lookup(:public_riiif_url)
       url = Addressable::URI.parse(CHF::Env.lookup(:public_riiif_url))
```
jrochkind: Maybe change these ENVs to public_iiif_url and internal_iiif_url too, with a switch on iiif_type or whatever?

lib/tasks/chf_tasks.rake
```diff
@@ -225,7 +225,7 @@ namespace :chf do
     end
 
     # Note this will not work on non-public images
-    desc 'ping riiif server to fetch all originals of publicly-visible images from fedora. `RAILS_ENV=production INTERNAL_RIIIF_URL=http://localhost bundle execrake chf:riiif:preload_originals`'
+    desc 'ping riiif server to fetch all originals of publicly-visible images from fedora. `RAILS_ENV=production INTERNAL_RIIIF_URL=http://localhost bundle exec rake chf:riiif:preload_originals`'
     task :preload_originals => :environment do
```
jrochkind: change comment to "iiif server" instead of "riiif server"? And above?

jrochkind: and change rake namespace to "iiif" instead of "riiif" I guess.

hackmastera:
> Can we make this code have fewer custom switches if we include more of the base path in the local_env.yml?

Yeah, I tried that but it didn't work with the implementation here https://github.com/chemheritage/chf-sufia/blob/riiif/app/helpers/riiif_helper.rb#L89

I figured once we choose where we're really going with this we can cut out the switches altogether, so I didn't worry about having them for now.

app/helpers/image_service_helper.rb
hackmastera: I stopped short of this so we weren't spending too much time changing our config requirements in local_env.

app/helpers/image_service_helper.rb
hackmastera: Right, it didn't seem like we were using them but I can add if you think we might.

lib/tasks/chf_tasks.rake
hackmastera: Will do this.

jrochkind:
> Yeah, I tried that but it didn't work with the implementation here 

> https://github.com/chemheritage/chf-sufia/blob/riiif/app/helpers/riiif_helper.rb#L89

I'm sure we can fix that somehow. I can look into it if you like. I suspect we just need to change the `path` passed in to start with `./` so it's a relative URL that can be added on to the base URL. 

I'm thinking we might end up leaving the switches in even in production. I think this stuff is gonna keep being tricky, and even once deployed to production we may run into temporary or permanent problems, the switches let us switch from one implementation to try another even in production to avoid downtime or whatever. 

Plus, I think if we can make our code just use straight IIIF api and be agnostic about what's backing it, which I think we can, we won't even need any switches I think!

hackmastera:
okay I'll work on it.

jrochkind:
I can work on it too if you want!

hackmastera:
No problem I'm on it! I am leaning toward requiring config for host and base path separately, because if you're using riiif on the same box as your app (e.g. dev) you want a relative url without a host.

jrochkind:
Even with riiif on same server on dev box an absolute base URL like `http://localhost:3000/image-server` works fine.  That's what I've been using. I think it's be better and less confusing to keep em together. It's sort of a coincidence that it happens to be on the same server in some cases, there's no real loss of efficiency or performance or anything to just using the full URL even if it could be a relative path.

PR #740, opened 2017-07-31
Bring rights_statement partial back in line on show page, fixes #727

### LINKED ISSUE #727: inline the rights partial in its other location
this will make it more straightforward, without the switch on `theme`. It's not really long enough to have as a partial just for readability I don't think, so if it's not going to be used somewhere else I'd rather inline it.

hackmastera:


jrochkind:
Fixes #727

PR #741, opened 2017-07-31
WIP imgix

jrochkind:
* Right now hard-coded to imgix, if use_image_server_on_viewer=true
* not yet used on show page or for jpeg downloads, just tiled viewer


jrochkind:
We are not going to continue with this at present.

PR #742, opened 2017-08-01
Generalize from `riiif` to `iiif`, allowing for use of cantaloupe or any iiif image server

hackmastera:
Show page verified working with riiif, cantaloupe, and neither (derivative thumbnails).
iiif preload verified working with riiif and cantaloupe.

Configuration now must be provided for a full url, including base path, e.g. (cantaloupe local_env.yml example) `iiif_public_url: //images.digital.chemheritage.org:8080/iiif/2` or (riiif ENV example) `IIIF_INTERNAL_URL=http://localhost:3000/image-service`

@jrochkind I believe this addresses all your notes from #739 except https://github.com/chemheritage/chf-sufia/pull/739/files/baf22ad92add4758e175220dce9adf766b7fe923#r130358339 since I couldn't think of a case where we would be using those. But if you think it's important I can do it or you can change it!

### DIFF
--- a/app/helpers/image_service_helper.rb
+++ b/app/helpers/image_service_helper.rb
@@ -47,9 +47,9 @@ module ImageServiceHelper
   # any responsiveness page layout. Sends somewhat more bytes when needed at some responsive
   # sizes, but way simpler to implement; keep from asking riiiif for even more varying resizes;
   # prob good enough.
-  def riiif_image_srcset_pixel_density(riiif_file_id, base_width, format: 'jpg', quality: 'default')
+  def riiif_image_srcset_pixel_density(file_id, base_width, format: 'jpg', quality: 'default')
     [1, BigDecimal.new('1.5'), 2, 3, 4].collect do |multiplier|
-      iiif_image_url(riiif_file_id, format: "jpg", size: "#{base_width * multiplier},") + " #{multiplier}x"
+      iiif_image_url(file_id, format: "jpg", size: "#{base_width * multiplier},") + " #{multiplier}x"
     end.join(", ")
   end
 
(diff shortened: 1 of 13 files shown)


app/helpers/image_service_helper.rb
```diff
@@ -82,13 +84,19 @@ def member_image_tag(parent_id:, member:, size_key: nil, lazy: false, use_image_
...
 
-  def create_riiif_url(path)
-    if CHF::Env.lookup(:public_riiif_url)
-      url = Addressable::URI.parse(CHF::Env.lookup(:public_riiif_url))
-      raise "public_riiif_url requires a valid URL with host, eg `http://host` or `//host`" if url.host.nil?
-      return Addressable::URI.join(url, path).to_s
-    else
-      return path
+  attr_accessor :public_host
+
```
jrochkind: what is an attr_accessor on a Helper module supposed to do? I think it's unneccesary and may do weird things?

lib/chf/utils/iiif_original_preloader.rb
```diff
@@ -0,0 +1,37 @@
...
+      def ping_to_preload
+        conn = Faraday.new(iiif_base)
+        conn.head ping_path
+      end
+
+      def ping_path
+        path_prefix = Addressable::URI.parse(iiif_base).path # may have terminal slash, may not
+        path_prefix << '/' if path_prefix[-1] != '/' # ensure terminal slash
+        path_prefix + "#{CGI.escape(file_id)}/info.json"
+      end
```
jrochkind: Add a terminal slash, but then append "/info.json" to it, resulting in a URL like `//info.json`?  Will prob still work with most servers, but. Maybe we should use Addressable here.

app/helpers/image_service_helper.rb
```diff
@@ -82,13 +84,19 @@ def member_image_tag(parent_id:, member:, size_key: nil, lazy: false, use_image_
...
+  def public_host
+    @public_host ||= begin
+      url = Addressable::URI.parse(CHF::Env.lookup(:iiif_public_url))
```
jrochkind: setting instance variable in a helper will set it on controller/viewcontext. If we need to do it, probably need to give it a less likely to conflict name like `@_image_service_helper_public_url`.  Or extract some of this to a helper class.

lib/chf/utils/iiif_original_preloader.rb
hackmastera: There's a file id between the path_prefix and the `/info.json`

hackmastera: from my messing around, Addressable doesn't manage slashes and frequently destroys portions of your path. if there is a way to use it for joins that would be great.

app/helpers/image_service_helper.rb
hackmastera: oh, oops, yeah I totally forgot I was in a helper :\

PR #744, opened 2017-08-03
image viewer, more robust errors

jrochkind:
For certain kinds of errors, a spinner was left forever, now we notice
and remove spinner, and add error message if needed. should make
debugging a little bit less confusing in some cases.


jrochkind:
@hackmastera approve before you leave? pretty simple one.

PR #747, opened 2017-08-07
use response.body=stream rails api to get quicker HTTP headers

jrochkind:
https://github.com/samvera/hydra-head/pull/421


jrochkind:
tagging @hackmastera because I'm merging while they are on vacation without review.

hackmastera:
I did look at this before I left actually I just failed to click the merge button. Thanks!

PR #749, opened 2017-08-10
Index representative_checksum too, for fingerprinted URLs

jrochkind:
Using the checksum in the URL for assets is nice, becuase it menas you can make
them infinitely cacheable (For instance for browsers), even if the image behind
a file_id changes (versions?).  I am doing this for DZI experiment, but it's
potentially applicable to all our image sources. 

But that means it's gotta be indexed, so you have it to generate the URLs
at display time. 

Yes, this apparently requires touching five different files!

### DIFF
--- a/app/presenters/chf/file_set_presenter.rb
+++ b/app/presenters/chf/file_set_presenter.rb
@@ -2,7 +2,7 @@ module CHF
   # Add some things we can use for our custom UI
   class FileSetPresenter < Sufia::FileSetPresenter
 
-    delegate :original_file_id, to: :solr_document
+    delegate :original_file_id, :original_checksum, to: :solr_document
 
     # Consistent API with GenericWorkShowPresenter, on show pages we often
     # only show permissoin badge if not open access.
@@ -18,6 +18,10 @@ module CHF
         Rails.logger.error "ERROR: Could not find FileSet #{id} original_file_id in Solr. You may need to reindex."
         return FileSet.find(id).original_file.id
       end
+    end
+
+    def representative_checksum
+      original_checksum
     end
 
     def representative_height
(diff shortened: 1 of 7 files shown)


app/presenters/chf/file_set_presenter.rb
```diff
@@ -20,6 +20,12 @@ def representative_file_id
       end
     end
 
+    def representative_checksum
+      # I can not explain why this didn't work with a delegate to solr_document
```
hackmastera: When you delegate to solr_document you still have to write an accessor method in the solr_document model. To bridge to the solr_name.

hackmastera:
Does this go along with a routing change? I will see more context in the dzi PR?

app/presenters/chf/file_set_presenter.rb
jrochkind: yeah, I figured out what was going on eventually! AFter I committed so didn't make it into my comments, I should fix them. 

Man, this stack.

jrochkind:
No routing change. In this case, it just changes the template for URLs in S3, those URLs are generated by ruby code in our app, Rails routing is not involved.

jrochkind:
Using this approach in other tile-back-ends may or may not require routing changes. I'm not really thinking those out, just going wtih DZI for now. This PR puts the checksums in the index so they _can_ be more easily used for file URI's, that's it.

hackmastera:
Got it, thanks!

app/presenters/chf/file_set_presenter.rb
hackmastera: Cool! Yeah, probably should fix the comment. Otherwise +1 to merge; feel free to go ahead!

PR #751, opened 2017-08-10
dzi

jrochkind:
Creating all tiles and pushing them to S3 with `rake chf:dzi:push_all` takes about 7 hours on staging, estimated about 9 on production. 

Things left to do

- [x]  mess with resque workers and roles, so the new box we set up for this purpose only does DZI's, and the old box does the old tasks (that may not be able to run on a separate server without shared FS,our new job can)

- [x] rake tasks to clean orphans in s3, or clear all. 

- [x] we can do some clever stuff in the JS to fall back to hydra-derivative thumb if OSD isn't available. Likely with a "Sorry, zooming not currently available" message.  That can actually work for canteloupe or any other server too, we can do this in master. 

- [X] Set at S3 HTTP caching headers. 

- [x] Make rake task to set CORS and any other setup on S3 bucket. 

- [x] Consider CDN. Individual tiles from S3 are coming back in like 40-70ms. Could a CDN make that even faster?. CDN could also gzip which would prob make it faster. (Or we could figure out if there's a way to deliver gzip'd response just from S3, probably by storing gzipped, possibly in addition to non-gzipped).  **decided ignore for now**

- [x] Note we have no auth at present. There are possibilities. Discuss. Currently all tiles are put in S3 such that they are publicly available. Maybe: No auth, document very clearly. investigate robots.txt on s3.  leave notes on plan for auth if we need it. **for now decided not to auth**

- [x] Mess with number-of-layers in DZI and/or other generation params to see if we can stop the "come in a block at a time" thing, and get it back to "come in blurry then get sharper" **updated openseadragon to latest for now**, that's pretty much all we can do without intensive debugging/enhancement of OSD. 

- [x] Ramelli show page is taking a long time to come in again. Did some refactoring changes I made effect performance? Or maybe OSD is being loaded too soon, and taking browser CPU time? Time for some more profiling. (Could not reproduce, guess is the app machine was just busy. all image requests go through rails app, which is unfortunate but not part of this issue). 

- [x] Docs on dzi setup:  rake tasks, env variables, auth options. 

- [ ] Infrastructure setup
   * For both production and staging, we need (so two of each of these, one prod one stage):
      1. An EC2 box for 'jobs', with app_role `jobs` set in local_env.yml
          * Ideally gets a hostnaem not just numeric IP to avoid confusion
          * needs `vips` installed, compiled so it can read TIFF and `dzsave`
          * will prob need GraphicsMagick in the future, so might as well if it's easy
          * The 'jobs' box does _not_ need redis or httpd/passenger running. Would leave more resources for jobs if they were not running. And does not need ports 80/443 open. 
          * The 'jobs' box needs `/opt/sufia-project/shared/config/redis.yml` set to proper redis server (which is currently app server) IP address, not 127.0.0.1 which is not accurate. Might be easier and avoid confusing to just do this on _all_ boxes, jobs server doens't really need special config if they all have a valid internal ip that can be reached?
      2. An S3 bucket. Perhaps `chf-dzi-prod` and `chf-dzi-stage`, matching the "service_level"/srv_lvl keys being used to avoid confusing. 
      3. AWS access key set with write permissions to the appropriate bucket (one prod one stage)
   * new local_env.yml keys. Values will be different in prod vs staging. While some of these keys will only be _used_ on certain 'roles', I think it makes sense just to give them to all machines for simplicity and consistency and not having to keep tweaking it if we discover we need a key on a role we didn't expect. (Our manual tweaks are pretty error-prone, best to do it and leave it). 
        * `aws_access_key_id:` $appropriate_id
        * `aws_secret_access_key:` $appropriate_key
        * `dzi_s3_bucket:` $appropriate_bucket_name
       * `image_server_on_viewer: dzi_s3`
  * Ideally this is all done in an automated fashion eg ansible. As we get more servers in our fleet, it is very easy for manual editing to result in inconsistencies between machines, resulting in time-consuming bugs. Minimizing _any_ manual editing would be great. 

- [ ] reindex on prod (can be done before merge, new index is in master)

- [ ] Create all DZI tiles on prod. I would probably deploy to prod first -- all viewers will show up with new "sorry, zooming not available" error, which is great, until the dzi's are created, then they'll start working as they are created. Then run `rake chf:dzi:push_all` on the new `jobs` server, properly env-configured as above.


jrochkind:
Jay gets a "HTTP error 0 can't load" from this URL: https://staging.digital.chemheritage.org/concern/generic_works/4b29b614k/viewer/6q182k19f

I'm having trouble getting that URL to load at all, weirdly. Error may not be specific to DZI.

sanfordd:
An error at
https://staging.digital.chemheritage.org/concern/generic_works/4b29b614k/viewer/6q182k19f
`Unable to open [object Object]: HTTP 0 attempting to load TileSource`
Was found by one check at that link, someone else checking did not encounter the error.

app/services/chf/create_dzi_service.rb
```diff
@@ -0,0 +1,263 @@
...
+          idletime:        60, # 1 minute. shouldn't matter, we have min and max the same
+          max_queue:       0, # unlimited
+          fallback_policy: :abort # shouldn't matter -- 0 max queue
+      )
+    end
+    self.thread_pool_executor # init now
+
+    # include the checksum so it's self-cache-busting if file at this URL
+    # changes, say, due to versioning. HOWEVER, this does make indexing
+    # somewhat slower. TODO optimize somehow. Less fetches to fedora and/or get
```
hackmastera: Remove this TODO? Not sure it's about this method.

app/services/chf/create_dzi_service.rb
```diff
@@ -0,0 +1,263 @@
...
+    # from solr.
+    def self.base_file_name(file_id, checksum)
+      CGI.escape "#{file_id}_checksum#{checksum}"
+    end
+
+    # returns [file_id, checksum]
+    def self.parse_dzi_file_name(base_file_name)
```
hackmastera: It's sort of confusing that this argument has the same name as the method

lib/tasks/chf_tasks.rake
```diff
@@ -214,6 +214,169 @@ namespace :chf do
...
+          progress.increment
+          progress.title = i
+        end
+      end
+      progress.title = i
+      progress.finish
+    end
+
+
+    # TODO, better API to call here, better performance, this is just WIP proof of concept
```
hackmastera: move to an issue?

lib/tasks/chf_tasks.rake
```diff
@@ -214,6 +214,169 @@ namespace :chf do
...
+        end
+        progress.increment
+        progress.title = "#{i} scanned, #{deleted.count} deleted"
+      end
+      progress.finish
+
+      # Have to do this a bit hackily, we'll actually iterate through every
+      # key, but the sdk #list_objects methods gets 'directories' out
+      # for us with #prefix
+      $stderr.puts "\nScanning for orphaned _files/ tiles...\n\n"
```
hackmastera: Interesting, have you observed this type of orphaned file? Any idea what would cause this?

app/models/chf/env.rb
```diff
@@ -139,18 +139,24 @@ def self.instance
...
-
-    define_key :use_image_server_on_viewer,
-      system_env_transform: BOOLEAN_TRANSFORM,
-      default: -> { false }
-
-    define_key :use_image_server_downloads,
-      system_env_transform: BOOLEAN_TRANSFORM,
-      default: -> { false }
-
+    # should be a recognized image service type, or nil/false for none, just hydra-derivatives
```
jrochkind: added link to code that has recognized values, to avoid comment becoming out of sync.

app/overrides/hydra/works/add_file_to_file_set_override.rb
```diff
@@ -0,0 +1,27 @@
+# Override and call super on #call to hook into "do this only after file is
+# actually added to fedora" Can't find a better way to do that.
+#
+# https://bibwild.wordpress.com/2017/07/11/on-hooking-into-sufiahyrax-after-file-has-been-uploaded/
+#
+# TODO: Test for uploading a new version of a file, do you still get this?
```
jrochkind: removed.

app/overrides/hydra/works/add_file_to_file_set_override.rb
```diff
@@ -0,0 +1,27 @@
...
+# TODO in dev does this end up prepended multiple times and calling super multiple times?
```
jrochkind: removed.

app/overrides/hydra/works/add_file_to_file_set_override.rb
```diff
@@ -0,0 +1,27 @@
...
+#      make that not so.
+
+Hydra::Works::AddFileToFileSet.class_eval do
+
+  AddFileToFileSetClassOverrides = Module.new do
+    def call(file_set, file, type, update_existing: true, versioning: true)
+      super.tap do
```
jrochkind: good call.

app/jobs/create_dzi_job.rb
```diff
@@ -0,0 +1,29 @@
+require 'concurrent'
+require 'aws-sdk'
+
+# Needs 'vips' installed.
+# Will overwrite if it's already there on S3 (or other fog destination)
+#
+#
+# s3 bucket needs CORS turned on! http://docs.aws.amazon.com/AmazonS3/latest/user-guide/add-cors-configuration.html
```
jrochkind: left it here, also linked from docs on rake task that does this for you.

app/jobs/create_dzi_job.rb
```diff
@@ -0,0 +1,29 @@
...
+#
+# TODO some cleverer concurrency stuff if two of these jobs try acting at the same
```
jrochkind: changed to POSSIBLE IMPROVEMENT

app/services/chf/create_dzi_service.rb
```diff
@@ -0,0 +1,263 @@
...
+  # self-cache-busting if the datastream at a file_id changes (versioning?). Does
+  # make the s3 keys kind of non-human-readable, but I think that's okay.
+  #
+  # Uses a bunch of Chf::Env settings:
+  # required:
+  #  * aws_access_key_id
+  #  * aws_secret_access_key
+  # optional (cause they have defaults):
+  #  * dzi_job_tmp_dir
+  #  * dzi_s3_bucket
```
jrochkind: fixed.

app/services/chf/create_dzi_service.rb
```diff
@@ -0,0 +1,263 @@
...
+  #  * dzi_s3_bucket_region
+  #
+  # Seek rake chf:dzi for tools for managing the S3 bucket contents.
+  #
+  # MAYBE would we better off using actual libvips bindings at https://github.com/jcupitt/ruby-vips
+  #   instead of shell out? I think it probably doesn't matter.
+  #
+  # TODO what do we want to do in dev? create dzi's to public/? No DZIs?
+  #
+  # TODO some cleverer concurrency stuff if two of these jobs try acting at the same
```
jrochkind: left some possible improvement/maybe that still apply.

app/services/chf/create_dzi_service.rb
```diff
@@ -0,0 +1,263 @@
...
+    self.vips_command = "vips"
+
+    class_attribute :jpeg_quality
+    self.jpeg_quality = "85"
+
+    class_attribute :suppress_vips_stderr
+    self.suppress_vips_stderr = true
+
+    class_attribute :cache_control
+    self.cache_control = "max-age=31536000"
```
jrochkind: That is literally copied from Amazon's suggestion. It's one year. will add comment.

app/services/chf/create_dzi_service.rb
```diff
@@ -0,0 +1,263 @@
...
+            open local_original_file_path, 'wb' do |io|
+              response.read_body do |chunk|
+                io.write chunk
+              end
+            end
+          end
+        end
+      end
+      Rails.logger.debug("#{self.class.name}: fetch_from_fedora: #{fedora_fetch_benchmark}")
+      unless response.code == "200"
```
jrochkind: not sure, but good call to be more defensive, changed to `unless response && response.code == "200"`

app/services/chf/create_dzi_service.rb
```diff
@@ -0,0 +1,263 @@
...
+
+    # uploading to s3 is actually the slowest part if done serially.
+    # We apply a healthy dose of concurrency.
+    def upload_to_s3!
+      s_time = Time.now
+
+      # All the jpgs, which are in a _files/ dir, and subdirs of that.
+      futures = []
+      dir_path = Pathname.new(local_dzi_dir_path)
+      path_prefix_re = /\A#{Regexp.quote(WORKING_DIR.end_with?('/') ? WORKING_DIR : WORKING_DIR + '/')}/
```
jrochkind: Added comment:       "We get back full paths from FS root from Dir.glob, need to change then to relative to working dir to get the S3 keys we want. "

app/services/chf/create_dzi_service.rb
jrochkind: removed. It may not be as optimized as it could be, but ti's just fighting with ActiveFedora and I don't even know.

app/services/chf/create_dzi_service.rb
jrochkind: I think it's intentional, normally what you'd pass in is indeed the same as what would be returned from the method. Make some sense?

lib/tasks/chf_tasks.rake
jrochkind: Yes, during my testing. If the job errors or is control-C cancelled after it's pushed some of the tiles but not all of them. Part of why I made sure the actual .dzi is pushed _last_, if it's pushed, you know they're all there, but if something errors in the middle, you may have partially uploaded tile set and no .dzi.

hackmastera: Makes sense! I noticed the way you did the .dzi last. The whole thing seems pretty solid!

jrochkind:
@sanfordd  uh oh. I think we don’t have resque workers running on jobs-prod.

`ps -ef | grep resque` gives me nothing on jobs-prod, something on jobs-staging.

i don’t know why those would be different if they are identical ansible-created! But we gotta get resque-pool running on jobs-prod.

jrochkind:
~~~
hydep@jobs-prod:/opt/sufia-project/current$ telnet 172.31.62.197 6379
Trying 172.31.62.197...
telnet: Unable to connect to remote host: Connection refused
~~~

~~~
 irb(main):001:0> redis = Redis.new(host: '172.31.62.197', port: '6379')
=> #<Redis client v3.3.2 for redis://172.31.62.197:6379/0>
irb(main):002:0> redis.get("foo")
Redis::CannotConnectError: Error connecting to Redis on 172.31.62.197:6379 (Errno::ECONNREFUSED)
~~~

PR #752, opened 2017-08-14
viewer has fallback url, with alert warning

jrochkind:
Fallback is the /downloads/$id?file=jpeg link as source. 

Tile-source-agnostic, just a viewer function.


hackmastera:
screen shot 2017-08-15 at 4 28 18 pm

PR #755, opened 2017-08-18
Allow physical container to be emptied; add feature spec to prevent regression

hackmastera:
Looks like whatever was preventing me from doing feature specs involving form submission is resolved. I must have made a config change somewhere along the way that fixed this!

This PR includes another feature spec fix which I did first just to see whether they were working.

### DIFF
--- a/app/forms/concerns/work_form_behavior.rb
+++ b/app/forms/concerns/work_form_behavior.rb
@@ -171,14 +171,15 @@ module WorkFormBehavior
       def self.encode_physical_container(params, clean_params)
         result = []
         CHF::Utils::ParseFields.physical_container_fields.values.each do |k|
-          # check for some data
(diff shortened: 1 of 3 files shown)


jrochkind:
failing CI?

jrochkind:
sometime before you leave, can you lead me through the special handling of physical container fields (that `join` stuff), and what it's motivation was?

hackmastera:
@jrochkind sure; send me a calendar invitation? You can call it "transfer of regrets"

jrochkind:
Ha, if it's a regret, we can probably change it back to 'normal' pretty easily I'd image. We'll see!

hackmastera:
😄 👍  I've made 1 or 2 issues for changing it but they haven't made it up the priority ladder. but yeah we'll talk about it!

PR #756, opened 2017-08-18
Only show 'my works' breadcrumb for logged-in users, fixes #746

### LINKED ISSUE #746: "my works" link visible when not logged in
public users viewing a work can see a link to "my works" in the breadcrumbs above the image on the work show page. Since they're not logged in, clicking on it just kicks them to a log-in screen. 

Can we get rid of this link for anyone not logged in?

hackmastera:


app/controllers/curation_concerns/generic_works_controller.rb
```diff
@@ -34,5 +34,11 @@ def t(key, interpolations = {})
     end
     helper HelperOverride
 
+    # Adds the 'My Works' breadcrumb; we only want this for logged-in users
+    # overrides https://github.com/samvera/sufia/blob/v7.3.1/app/controllers/concerns/sufia/works_controller_behavior.rb#L93
+    def add_breadcrumb_for_controller
+      super if current_ability.current_user.staff?
+    end
```
jrochkind: would it make more sense just to do `current_ability.current_user`?  If we had a logged in user that was not staff, they should maybe still get that?

hackmastera: `staff?` is the same as logged-in in our application. We wrote `staff?` to be a short way to check whether someone is logged in. A non-logged in user still has current_user because there's a dummy / non-persisted user that's used kind of as a nil object.

https://github.com/chemheritage/chf-sufia/blob/master/app/models/user.rb#L34

jrochkind: Yes, it's the same right now. But it could change in the future. 

While "staff?" and "logged in" are the same thing right now, in some cases we _really_ want to be checking for "staff" (if there was a logged in user that was not staff, they should not be included). 

But in this case I think we _really_ want to check for "logged in" -- if there's a logged in user that's not staff, they _should_ still get the breadcrumbs.

Do you happen to know if there's an easy way to distinguish that dummy user?  If there is, I _think_ that would be better semantically. But not really a big deal, if it's a confusing pain this is probably good enough for now -- hypothetically adding non-staff logged in users in the future would prob break lots of things already.

hackmastera: @jrochkind we could change the `staff?` method I linked to above to `logged_in?` and then alias `staff?` back to it, changing this to use `logged_in?` would that work?

That method uses a check for belonging in the group `registered`. You can also check `persisted?` on the user; the dummy object returns `false`.

jrochkind: I would either check `current_ability.current_user.persisted?` here, or create a logged_in? cover method that just does that (`persisted?`) -- **or** just leave it how it is now. Your call.

hackmastera: @jrochkind okay what do you think of it this way?

jrochkind:
works for me! merge once CI is done.

PR #758, opened 2017-08-18
Fix show page for work with more than one member

hackmastera:
@jrochkind I think this got overlooked in some refactoring. Please take a close look and make sure this matches your understanding of the current state of the codebase!


jrochkind:
weird, maybe a merge gone wrong. Looks right I think. You tested and it works?

hackmastera:
Yes I tried it. Before this change I could not load a show page for a work with more than one member. After this change I could.

PR #773, opened 2017-08-23
Set default search fields to the same as 'all fields', fixes #673

### LINKED ISSUE #673: Collections search not searching all fields
The use case is that a researcher wants to know everything that was digitized in a Series or Box in a Collection (i.e. Beckman Historical Collection). They can search for "Box 50" or "Series X" in the general search and get hits, but if they go onto the Collection page and search just that Collection, they get no results. We haven't tested this for all fields, but we're wondering if the same metadata fields were indexed in the Collections search as were for the general search.

hackmastera:
@jrochkind any reason why our default search fields shouldn't be the same as our 'all fields' keyword search?

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -86,7 +86,7 @@ class CatalogController < ApplicationController
     config.default_solr_params = {
       qt: "search",
       rows: 10,
-      qf: "title_tesim name_tesim"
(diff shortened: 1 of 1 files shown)


jrochkind:
makes sense to me. No idea in what case this 'default' search would be triggered vs our single configured search field. It's possible the 'default' search never happens. I dunno! But I don't see any problems.

jrochkind:
Fixes #673

PR #774, opened 2017-08-23
Remove AddCatluAccess task

hackmastera:
We don't need this sort of task anymore, since we have admin users


jrochkind:
nice, thanks!

Issue #562, opened 2017-05-30
Change Genre term 'Diagrams' to 'Charts, diagrams, etc.'


catlu:
New URI: http://id.loc.gov/authorities/subjects/sh99001425.html

Will need to update existing data as well, right? Should I make another ticket for that?

hackmastera:
Yes this will require data migration. No need to make another ticket.

hackmastera:
I believe we can do this 100% risk-free with no downtime as follows:

1. Delete the 'Diagrams' term from the list of options
2. Run the data migration task
3. Add the new 'Charts...' term to the list.

I'll test it on dev. If it works, we should notify staff users before we do it since they'll see the term disappear from the form.

hackmastera:
An extra safeguard would be to check for 'Diagrams' on save and change it to the new term so that anyone who uses it after the task has run, but before the option is removed, would get the correct term.

hackmastera:
Steps:
- [ ] Test process on staging
- [ ] Merge #775, deploy
- [ ] Run data migration task (`chf:data_fix:genre_charts`)
- [ ] Merge #776, deploy

PR #775, opened 2017-08-24
Charts genre step 1

hackmastera:
Pending testing on staging.

This PR:
- Adds the new option to the form.
- Adds the migration task.
- Adds a data filter to form processing that catches any attempts to save the old term and converts it into the new term. This will keep new edits from creeping in after the task has run but before the old option has been removed from the form. This code is intended to be temporary, removed in the follow-on PR. Thus it is isolated in its own commit.

The removal of the old option is in a follow-on PR

Refs #562


hackmastera:
successful test on staging; ready for merge

PR #780, opened 2017-08-24
make sure honeybadger reports better env for deploy tracking

jrochkind:


hackmastera:
@jrochkind how does the cap configuration relate to honeybadger? Not exactly sure what's going on here; maybe a comment with the configuration would be useful to some future person as well?

jrochkind:
Honeybadger also, unless you turn it off, tracks deployments to give you a log of what was deployed when/where. Which is kinda helpful.

http://docs.honeybadger.io/ruby/getting-started/tracking-deployments.html

https://app.honeybadger.io/projects/53196/deploys

hackmastera:
Very nice!

Issue #791, opened 2017-08-28
keyboard control issues


jrochkind:
* accessible keyboard controls, can't hit return to 'click' when thumb is focused on show page
* viewer keyboard controls not working on FF
* viewer keyboard controls sometimes stop working when 'viewer' loses focus.

PR #805, opened 2017-08-28
viewer, various keyboard control fixes

### LINKED ISSUE #694: keyboard shortcuts don't work when popup is active
Either it should be clear that it needs to be closed to start using them, or they should work when it's open.

jrochkind:
closes #694 and #791

### DIFF
--- a/app/assets/javascripts/chf_image_viewer.js
+++ b/app/assets/javascripts/chf_image_viewer.js
@@ -426,11 +426,6 @@ ChfImageViewer.prototype.initOpenSeadragon = function() {
 };
 
 ChfImageViewer.prototype.hideUiElement = function(element) {
-  if (document.activeElement == element) {
-    // If it was focused and we're hiding it, make sure to switch focus
-    // to modal, so keyboard shortcuts and tab still works right.
-    this.modal.focus();
-  }
   element.style.display = "none";
 };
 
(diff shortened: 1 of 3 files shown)


jrochkind:
I thought I was done before, but now I think I'm done again! Tested on IE11.

app/assets/javascripts/chf_image_viewer.js
```diff
@@ -234,123 +243,116 @@ ChfImageViewer.prototype.locationWithNewPath = function(newPath) {
 };
 
 ChfImageViewer.prototype.onKeyDown = function(event) {
-  if (this.dropdownVisible) {
```
hackmastera: It seems like you're removing all references to this? Maybe you can delete these, also https://github.com/chemheritage/chf-sufia/blob/7eeb32669b4395042fdb7c50c1d148fde2642c63/app/assets/javascripts/chf_image_viewer.js#L479-L487

jrochkind: yeah, I probably could, cause you're right, I did, good catch. But I sorta feel like leaving it there in case I need it later, this stuff has been kind of tricky to get right!

hackmastera: 👍  works for me; I'll merge.

hackmastera: haha i was too late

PR #815, opened 2017-08-29
rake dzi_push make our hacky way of using rake args work

jrochkind:


hackmastera:
Not sure what is going on with travis...

Issue #709, opened 2017-07-19
Add "Search Only Public Domain" checkbox under search


MDiMeo:

hackmastera:
@MDiMeo I've got something ready for us to look at; I think we need to talk more about this and #796 before I move forward -- for Tuesday's meeting.

hackmastera:
fixed in #818

PR #818, opened 2017-08-29
Add checkbox for public domain facet to search, refs #709

hackmastera:


jrochkind:
nice use of search builder!

Issue #793, opened 2017-08-28
slowness on /viewer links on large show page


jrochkind:
If linking directly to a /viewer link on a show page, and the show page has many thumbs (like ramelli), there are 5+ seconds when JS is frozen (spinners aren't even spinner), and viewer isn't triggering. Could be aspectratio related.

Issue #792, opened 2017-08-28
aspect ratio issues


jrochkind:
* aspectratio not working right for work-as-member thumbs
* aspectratio not being triggered on viewer

jrochkind:
Okay, so we use the [lazysizes aspectratio](https://github.com/aFarkas/lazysizes/tree/gh-pages/plugins/aspectratio), to try to set not-yet-loaded-image spaces to be the right size for the image once loaded. It works whether you are lazy loading images with lazysizes, or loading them the ordinary way (in case they are slow to load even using standard `<img src`). 

We use this on the show page, and for the viewer thumbs, both of which do use lazysizes. I think it's _mostly_ working on the show page (although there may be problems there). But there are more severe problems on the viewer thumb, on many-membered works. 

The viewer thumbs are sometimes not actually succesfully aspect-ratio-sized. I think maybe when the page loads (with viewer hidden), there isn't enough info for aspectratio to size the images that are currently hidden, so when you do open the viewer, they are not sized. 

But _worse_, if you try to use a `/viewer` link for immediate automatic opening of viewer. The browser kind of locks up with too much JS trying to do too much. I think maybe the fact that aspectratio code is triggered _while the viewer modal animation is in progress_ confuses code or otherwise puts too much stress on things. 

But I tried various methods for delaying the aspectratio code until after the viewer was fully shown, which _did_ improve things, but still ended up locking up the browser, especially in non-Chrome browsers. 

I think I will abandon automatic aspectratio js code on the viewer thumbs. Instead, I will, on the back-end, set the HTML `height` attribute based on known aspect ratio. In the ERB, no JS. 

This will mean that viewer thumb images will have a fixed width whether they are in left-sidebar, or in smaller-screen bottom-bar. Before I was trying to do clever things to fix the width but let the height vary in sidebar, but fix the height and let the width vary in bottombar. Those clever things may have contributed to the problem, and are no longer possible with this new approach. 

Hoping aspectratio JS on _show_ page works okay once it's removed from viewer thumbs. 

@hackmastera fyi

PR #820, opened 2017-08-30
Improvements to lazyloading images and aspectratios

jrochkind:
See individual commits. Includes removing use of js lazy aspectratios
in viewer thumbs, for hard-coded ones instead. 

I believe lazyloading with aspectratios should be working well here,
and no longer will a direct to /viewer link on a many-membered work
freeze up the browser for a while. 

One thing we might want to do that I have NOT done yet is replace
the animated spinner with a static wait/progress icon of some kind. 
I think so many instances of that animated spinner on a page are putting
stress on CPU. 

Ref #792 #793

### DIFF
--- a/app/assets/stylesheets/local/chf_image_viewer.scss
+++ b/app/assets/stylesheets/local/chf_image_viewer.scss
@@ -295,6 +295,7 @@ $chf_image_viewer_thumb_width: 60px; // Needs to match MemberHelper::VIEWER_THUM
       &.lazyload, &.lazyloading, &.lazyload {
         background: #AAA asset-url('spinner.svg') no-repeat center;
         background-size: 34px 34px;
+        background-clip: content-box;
       }
     }
   }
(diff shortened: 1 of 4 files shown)


jrochkind:
This is deployed to staging too right now I think.

jrochkind:
@hackmastera So previously, responsive CSS was used to make it so:

When media query put the thumbs on the sidebar (non-tiny-screens), thumb images were limited to 60px wide (so there could be two of them side by side), with height for whatever the aspect ratio of the image was. 

When media query put the thumbs on the bottom (tiny screen), thumb images had constrained _height_ (so they'd fit in the horizontal bar), and allowed the width to be as wide or as narrow as it needed to be for aspect ratio. 

This was clever, but too clever. Some of the things involved in doing that just weren't working well, mainly the JS aspectratios thing, for reasons I couldn't entirely figure out. 

So, now, the images are just set to 60px wide with height appropriate for aspect ratio, period, and stay that way on responsive small-screen thumb-bar-on-bottom too.

hackmastera:
Not much useful for me to say here; looks like you are either some kind of magician or you did a bunch of trial-and-error css development which I won't fully understand because I wasn't there (I also don't think I really observed the bugs). But nothing jumps out as weird to me and I'm glad you got it working!

PR #828, opened 2017-08-31
New home page text

### LINKED ISSUE #759: introductory text not informative enough
Users confused about What CHF is. Users confused about where these objects come from. Users don't know what to expect when they submit a search.

jrochkind:
Closes #759


jrochkind:
thanks fixed missing link

PR #830, opened 2017-08-31
Automatic slack deploy notifications

jrochkind:


lib/chf/slackistrano_messaging.rb
```diff
@@ -0,0 +1,20 @@
+module Chf
+  class SlackistranoMessaging < Slackistrano::Messaging::Base
+    def channels_for(action)
+      channels = if fetch(:stage).to_s == "production"
+        ["#digital-general"]
+      else
+        ["#{}digital-technical"]
```
hackmastera: What do these braces mean? It seems like that would negate the presence of the `#`.

jrochkind: bug/typo, yup. Wonder how it was working anyway? Will remove.

hackmastera: Editor being too clever?

jrochkind:
yup it was editor being too clever.

PR #840, opened 2017-09-05
Clean up and better branding of facets sidebar.

jrochkind:
PR into visual_ux_refresh branch


hackmastera:
@jrochkind looks good to me. Are you checking multiple browsers as you go, or waiting until the end?

jrochkind:
I haven't been checking multiple browsers, good reminder. I would be shocked if any of this relatively simple styling is different in different browsers, but I've been shocked before. :) I think this one I'm fine not checking yet, and checking when I have more of search results page done.

PR #849, opened 2017-09-08
Upgrade sufia to 7.4 and rails to 5.0.6

### LINKED ISSUE #545: Upgrade to Rails 5.0
Before soft launch, if we have time, me and @hackmastera think it might make sense to upgrade to rails 5.  Prob not 5.1 yet. 

Rails 4.2 which we are on is effectively end-of-lifed, and doing it sooner would give us the opportunity to write code now that will work in rails5 without deprecation using new rails5 api. 

@hackmastera thinks it should be fairly straightforward based on our hyrax experiment (we are _not_ suggesting upgrading to hyrax any time soon, before soft launch). 

@MDiMeo fyi

### LINKED ISSUE #786: release a sufia version that will allow dependency upgrades
we're on 0.11.2; current version is 1.1.0.

changelog https://github.com/carrierwaveuploader/carrierwave/blob/master/CHANGELOG.md
hyrax update (doesn't look like there were any changes necessary) https://github.com/samvera/hyrax/commit/773a80db98034765c34e5987d8525736fc32641b

hackmastera:
fixes #786 
fixes #545 

Looks okay on dev. Deployed to staging for more comprehensive testing.


hackmastera:
Status:

There appear to be loading issues when deployed to staging, possibly due to http://guides.rubyonrails.org/upgrading_ruby_on_rails.html#autoloading-is-disabled-after-booting-in-the-production-environment. See also http://blog.bigbinary.com/2016/08/29/rails-5-disables-autoloading-after-booting-the-app-in-production.html

Also, there are some deprecation warnings.

And a couple of failing tests. One looks to be related to the loading problems. The others look routine; actually maybe I can fix them now.

hackmastera:
Fixed a bunch of tests due to deprecated `assigns`.

I think this is almost done, @jrochkind! I hope you can find a minute to figure out the loading-in-production issues. I think now is the time to wrap up these upgrades, since we have momentum off of the sufia release. Sorry I couldn't get it all nice and finished up for you!!

app/controllers/catalog_controller.rb
```diff
@@ -6,7 +6,6 @@ class CatalogController < ApplicationController
 
   # These before_filters apply the hydra access controls
   before_filter :enforce_show_permissions, only: :show
-  skip_before_filter :default_html_head
```
jrochkind: ??

Gemfile
```diff
@@ -1,11 +1,15 @@
...
 gem 'blacklight_range_limit', ">= 6.1.2"
 gem 'riiif', '~> 1.4'
 # It may seem like this is a dependency upstream, but we need it here.
 # https://project-hydra.slackarchive.io/dev/page-100/ts-1491587437169731
+# Update: slackarchive permalink doesn't work after some max. search for 'rsolr' and look
+# at April 7 results for hints of the conversation? Unfortunately you can't get context.
+# I tried relaxing this with the sufia 7.4 upgrade but it a bug where submitting a record
+# edit form did not correctly load the redirect back to the show page.
 gem 'rsolr', '~> 1.0'
 gem 'active-fedora', '~>11.1.6'
```
jrochkind: sufia will allow rsolr 2.0, but we found it actually only worked with rsolr 1.x, is that right?

app/controllers/catalog_controller.rb
hackmastera: that method no longer exists. I believe it was generated code.

jrochkind: thank you!

Gemfile
hackmastera: I don't really know. I had pasted a slackarchive link here, but it no longer works and as I wrote in the comment I can't get to the context around the discussion.

I do know that I tried allowing rsolr > 1.0 and observed the bug described, which went away when I replaced the rsolr line.

I didn't investigate farther, figured we didn't need to upgrade rsolr to upgrade rails, which was the primary goal of this work.

jrochkind: Okay, thanks. It's distressing becuase there's another commit in Sufia where @mjgiarlo is specifically making it work _with_ rsolr 2.0.  Maybe it's something specific to our app rather than sufia "7.4" that makes it not. Hard to say. 

I think really what we're learning is that sufia is not only not maintained, but essentially not maintainable, things that seem like they should be simple won't be, we probably should never try to update sufia again, just live with it as it is, or painfully upgrade to hyrax (or abandon it). I ran into a similar thing trying to do a [simple dependency change on curation_concerns](https://github.com/samvera/curation_concerns/pull/1186), and then basically felt like giving up.

jrochkind:
updating a whole bunch of dependencies. Deps that broke the build or that I was feeling unconfident about upgrading I didn't, but upgraded a bunch of stuff, as long as we're upgrading and manual testing.

jrochkind:
Also curation_concerns to 1.7.8 in order to allow kaminari 1.x.

Issue #763, opened 2017-08-21
Rewrite 'about' and 'FAQ' content


hackmastera:
Several users read the About page and FAQ first, but found them boring or too technical. Didn’t feel the content was written for them and didn’t care about the Collections policy. Links leaving the Digital Collections site confused at least one user.

MDiMeo:
Consider headings for Technical Details and Acknowledgements.

MDiMeo:
This is the Stanford example of the more technical About page that I was thinking of in today's meeting. https://bv.stanford.edu/en/about I see they have headings and a secondary navigation, but the introductory test is more user friendly and general.

MDiMeo:
I rewrote the About page and had it approved by Communications. Please delete the current text and replace it with this: 

**About**
The Chemical Heritage Foundation’s Digital Collections offer highlights from our library, archives, and museum collections. Our chronology ranges from the Roman Empire to the 21st century, and the materials include everything from fine art to rare books and from scientific objects to photographs. 
The purpose of CHF’s Digital Collections is to manage, preserve, and provide access to our digital assets all in one location. Although the Digital Collections include only a small portion of our entire collection, new material is added every day. See our [Collections page](https://www.chemheritage.org/collections) for more information about our library, archives, and museum. 


**Copyright**
Possession of a digital reproduction of CHF’s collection materials does not constitute permission to publish, exhibit, or distribute. In most instances CHF does not hold the copyright to its materials and cannot grant or deny permission to use them. 

CHF staff makes every effort to inform patrons of copyrights and other restrictions that may apply to the works by providing a link to RightsStatements.org or Creative Commons in the Rights field in the object’s metadata record. Items we believe to be free of copyright and in the public domain are clearly identified. However, patrons are solely responsible for determining the copyright status of any materials they wish to use, investigating the owner of the copyright, and obtaining permission for the intended use.

The U.S. Copyright Law contains an exception for the so-called fair use of copyrighted materials for the purposes of teaching, scholarship, research, criticism, commentary, and news reporting. Patrons are solely responsible for determining whether their intended use of a copyrighted work is fair and for responding to any claims of copyright infringement that may arise from use of the material. For more information on copyright and fair use, please refer to the [U.S. Copyright Office website](https://www.copyright.gov/).


**Technical Summary**
[...]

PR #873, opened 2017-10-02
New About page text

jrochkind:
Ref #763


jrochkind:
do over, something weird in teh pull

PR #881, opened 2017-10-04
[waiting if we need it] patch activefedora to support timeout config

jrochkind:
Would need edit to fedora.yml to add a timeout value, which I think is ansible-controlled

### DIFF
--- a/app/overrides/active_fedora/fedora_override.rb
+++ b/app/overrides/active_fedora/fedora_override.rb
@@ -1,0 +1,27 @@
+# Override to add a timeout option to fedora.yml config, so we can increase it.
+# Should be in some future ActiveFedora after 11.4.0.
+# https://github.com/samvera/active_fedora/issues/1105
+
+# timeout value in fedora.yml is in SECONDS
+
+if Gem.loaded_specs["active-fedora"].version >= Gem::Version.new('11.4.0')
+   msg = "
+   Please check and make sure this patch is still needed\
+  at #{__FILE__}:#{__LINE__}\n\n"
+   $stderr.puts msg
+   Rails.logger.warn msg
+end
+
+module ActiveFedoraOverride
+  def authorized_connection
+    super.tap do |conn|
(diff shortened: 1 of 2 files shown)


jrochkind:
See also/instead

Chris Colvard [2 minutes ago] 
This is a monkey-patch that I haven’t gotten around to proving out for Avalon but it _might_ work for you:
https://gist.github.com/cjcolvar/b4671e6d7187a55065c5b4ea57e09d1c


Chris Colvard [2 minutes ago] 
based upon the conversation here: https://jira.duraspace.org/browse/FCREPO-2544

jrochkind:
Or maybe not. 

Ben Armintor [4:28 PM] 
@cjcolvar I'm not sure how the rest of AF would react to ntriples, but in the particular case of the root node - which you probably don't have other properties on? - it seems like it would be wise to use ntriples anyway


[4:28] 
that is, even if you backed off of that Accept header for the actual objects, there's probably no reason not to use ntriples for listing the root node.


Jonathan Rochkind [4:28 PM] 
@barmintor @cjcolvar that monkey patch doesn’t just apply to the root node request though, I think, but everything?


[4:29] 
Monkey patching ot somehow only use ntriples for the root node request seems a lot harder.


Ben Armintor [4:29 PM] 
@jrochkind yes, that's why I'm suggesting a more contextual approach


Jonathan Rochkind [4:29 PM] 
okay, cool. That warning is enough to make me NOT use @cjcolvar’s code snippet. Modifying AF to do that is going to be a lot harder, and a mess via monkey patching I think.

jrochkind:
maybe later

jrochkind:
(we may now be using a more recent active-fedora that has a settable timeout value)

jrochkind:
FWIW we are not on a version of active-fedora that includes timeout support. would have to upgrade to active-fedora 12.x. 

Or use this runtime patch. 

https://github.com/samvera/active_fedora/commit/a58168680214f8a93ff7e46e9b8b21bb4939ebad

https://github.com/samvera/active_fedora/issues/1105

Issue #761, opened 2017-08-21
50% of users tested didn't scroll down front page


hackmastera:
- Decreasing size of the hero image would make it more likely that users would see content below.
- Color of 2nd content area same as color of footer. Would a different color help?
- 2nd content area has no header. Would adding a header help this? Would probably be good, regardless. "Popular Topics"?
- Subnavigation that pops people around the page?

jrochkind:
Why do we want more percent to scroll down?

MDiMeo:
This is a good question worth discussing. We spent time building curated selections to showcase collection highlights and reinforce the institution's brand, and I think the content below the fold does convey some of our strengths to the user. Many people reported that they felt confused by the scope and purpose of the site, and they went to the About and FAQ pages to try finding that information. Redesigning the page slightly to make sure they see some of this other content might help answer some of these questions, especially for the casual or first-time user. On the other hand, there were some users who just started by putting a search term relevant to them in the box and searched. These users who have a clear task in mind would probably not have used browse as an entry point into collections in the first place, so I'm less worried about them not having scrolled down. But I do think this could serve as a possible entry point for some of the other behaviors we identified in the user personas (and confirmed with usability testing.)

jrochkind:
Hmm. This makes me think maybe we _aren't_ that happy with the home page design we got from 2010, after user-testing, and a complete home page redesign is needed, rather than just trying to tweak it. Maybe.

hackmastera:
@jrochkind yeah I wondered the same thing; seems like it's worth considering.

hackmastera:
Example: NYPL site changes the hero image aspect (without resizing) if you resize your window vertically https://digitalcollections.nypl.org/

jrochkind:
With #905 done, do you want to close this, move it, or what, @MDiMeo ?

jrochkind:
@MDiMeo what do you want to do with this ticket?

MDiMeo:
Yes, this works for now. Let's close it - thanks!

PR #905, opened 2017-11-21
Smaller front-page hero

jrochkind:
With changes and crazy arithmatic to try to keep whitespace and alignment looking good at all
screen sizes.

Ref #761

### DIFF
--- a/app/assets/stylesheets/local/homepage.scss
+++ b/app/assets/stylesheets/local/homepage.scss
@@ -48,21 +48,30 @@
   .search-container {
     padding-left: 5px;
     padding-right: 5px;
-    @media (max-width: $screen-md-min - 1) {
-      padding-bottom: 30px;
-    }
+  }
 
-    @media (min-width: $screen-xs-min) and (max-width: $screen-xs-max) {
-      margin-top: 30px;
-    }
-    @media (min-width: $screen-sm-min) and (max-width: $screen-sm-max) {
-      margin: 30px 20px 0;
-    }
-    @media (min-width: $screen-md-min) {
-      margin: 10% 15% 0 8%;
-    }
-    @media (min-width: $screen-lg-min) {
-      margin: 20% 15% 0 8%;
+   @media (min-width: $screen-md-min) {
+    .home-right-column {
+      // crazy way to keep .search-container vertically centered
+      $hero-aspect-ratio: 1.366; // w divided by height of hero image in left column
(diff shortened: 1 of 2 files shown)


jrochkind:
@MDiMeo I tried to make the image smaller while changing as little else as possible. Still had to change a bunch of other things and do some weird arithmatic to try to keep the white space and alignment looking good at any screen size, and keep the content from getting too large and making white space under the hero image at some screen sizes. 

I can demo it on staging when you are back. 

Here is a screenshot at a very large screen size:

very big screen

Here is a screenshot at the smallest screen size before it wraps to one column:

just before wrap

Is this basically what you were imagining?

jrochkind:
@MDiMeo a 50% screen width version is currently live on Staging.

jrochkind:
Per discussion, if you want a different aspect ratio, you or someone else can certainly do a crop with the aspect ratio you want and we can try that.

MDiMeo:
@jrochkind Thanks! I mostly like it at 50%. My only issue is that it has significant white space under the hero image that I think has the unintended result of making the page look less professional. Can you see that on your screen or do you want a screen shot of what I see when I load it? And would that be a huge pain to eliminate that white space?

jrochkind:
Screen shot _under_ the hero image?  Yes, I don't see that, screen shot please.

MDiMeo:
staging screenshot
There was white space on both of @catlu's monitors, too. We both agreed that if the white space was removed that the 50% screen width would be a slight improvement on what we currently have.

jrochkind:
OK, thanks, I'm having trouble reproducing that. Can you tell me what browser you are using, and if that white line appears at all sizes as you resize your browser window, or only at certain ones?

MDiMeo:
I have been using Chrome and yes, it stays the same size as I resize my window. But I just tried it in IE and don't see it there.

jrochkind:
Odd, I don't see it in MacOS Chrome, Safari, or Firefox; or Windows IE11. 

I forget, are you on Windows Chrome?

MDiMeo:
Yes. And so is Cat. We can ask @sanfordd to check his when he's back from lunch. I'm signing off for a short bit now, too.

jrochkind:
I successfully reproduced the problem on Windows Chrome, okay!  Wow, it's been a while since i saw a difference between a browser on Windows and Mac, that doesn't really happen anymore! But is here! 

Okay, when I get that fixed, shall I merge to master and deploy to production?

jrochkind:
OK, with white space fixed, merging and deploying to prod.

PR #976, opened 2018-02-06
WIP Expanded search tools on Collections pages

jrochkind:


jrochkind:
Abandoning this approach for #983

PR #986, opened 2018-02-20
RIS export

### LINKED ISSUE #565: Fix citation export tools (with RIS export)
To be determined/discussed further:

None of these have been examined to make sure they work with our customized metadata fields. If we aren't able to do this before soft launch, consider removing them until we're able to test / fix them.

maybe on-screen human readable as well as fix endnote, zotero, export. (what do we do with makers? get some samples of how things ought to be cited created by a human). Citation button on viewer?

jrochkind:
Closes #565

We probably want to create new tickets fori (@mdimeo):
* Looking into improving Zotero extension "toolbar" button
* Looking into improving "child work" issue (including current cases where there is no parent in the system for the 'child')

### DIFF
--- a/app/assets/javascripts/popovers.js
+++ b/app/assets/javascripts/popovers.js
@@ -1,0 +1,3 @@
+$( document ).ready(function() {
+  $('[data-toggle="popover"]').popover()
+});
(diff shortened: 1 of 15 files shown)


jrochkind:
A thread in where I ask some questions about Zotero imports: https://forums.zotero.org/discussion/comment/301945/

PR #991, opened 2018-02-21
correct cap deploy url to new github org sciencehistory

jrochkind:


jrochkind:
I had some inkling that this might not actually work for capistrano, simply changing the `repo_url` on an already existing deployment. 

However, searching reveals that it didn't used to work, but ought to since Cap 3.7.2. We are are on 3.9.1, so we _ought_ to be good. 

https://github.com/capistrano/capistrano/commit/109c2d205ea7b8449aa8e0c391cdaf05a5a0c0db

PR #1025, opened 2018-03-19
Feature to log out admin logins while leaving public app up

### LINKED ISSUE #1014: add feature to lock out staff for maintenance (without shutting down public interface)

jrochkind:
Implemented super cheesy hacky. 

1. Hook into a devise feature meant for user 'activation' to let all users not login
2. be even meaner to devise by using some hacky ruby/rails craziness to tell it not to present
   the usual login screen if we have logins disabled. 

If we set logins to be disabled, and a logged in user accesses the app, they will be immediately logged out. 
If they were trying to do something that required a login, they'll be redirected to the 'login' screen, which
has been replaced with a message actually saying "Sorry, logins disabled."

They'll have to login again when logins are no longer disabled. 

If they never accessed the app when logins were disabled, they'll find themselves still logged in after the window though. But if they acccess during window, they'll get logged out and have to log in again.

Closes #1014

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -35,6 +35,23 @@ class User < ApplicationRecord
     groups.include? 'registered'
   end
 
+  # A devise method, we're taking advantage of it for a cheesy way to lock out
+  # admin users based on CHF::Env variable.
+  def active_for_authentication?
+    if CHF::Env.lookup(:logins_disabled)
+      false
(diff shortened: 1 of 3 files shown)


eddierubeiz:
Eddie here. Just wanted to signal that I did look at the changes you made.

PR #1029, opened 2018-03-23
Update label before index

eddierubeiz:
Here's a naive attempt to fix https://github.com/sciencehistory/chf-sufia/issues/428. Mostly submitting this as an exercise; I don't claim to know exactly what I'm doing.

### DIFF
--- a/app/models/generic_work.rb
+++ b/app/models/generic_work.rb
@@ -28,7 +28,7 @@ class GenericWork < ActiveFedora::Base
   # representative_height, representative_original_file_id, and representative_checksum
   def update_index(*args)
 
-    # Thisgit diff fixes bug https://github.com/sciencehistory/chf-sufia/issues/428 .
+    # This fixes bug https://github.com/sciencehistory/chf-sufia/issues/428 .
     # It updates the display_label property for each inscription associated with this work
     # before said display_label is indexed into SOLR's 'inscription_tesim' field.
     # That field, in turn, is what is shown on the individual work display page
--- a/app/models/generic_work.rb
+++ b/app/models/generic_work.rb
@@ -27,6 +27,7 @@ class GenericWork < ActiveFedora::Base
   # This code goes with custom code in indexer to index representative_width,
   # representative_height, representative_original_file_id, and representative_checksum
   def update_index(*args)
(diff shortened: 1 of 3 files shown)


jrochkind:
(Nice job figuring out a solution!)

jrochkind:
It occurs to me that:

1. A comment by that one-liner pointing to the URL for this issue might be good, cause it's mysterious from the code what it is or why it matters. 

2. An automated test would be great. I'm not a total stickler for automated tests, especially when they are painful to do in a given situation and especially when the thing tested is about fighting with sufia, but maybe it would be a good learning experience?   If so, I might try to do it as a front-end test, modelled after this one (although this one only tests creating new work, not editing an existing one). https://github.com/sciencehistory/chf-sufia/blob/master/spec/features/new_work_feature_spec.rb

eddierubeiz:
Good call. I ended up modifying and extending the test you mention in #2 rather than splitting off a new test, since the existing test functions perfectly as the setup for the new test I need. I'm totally open to splitting it off into its own separate test, which would be more elegant but (I suspect) somewhat slower. But it works great as a regression test for bug 428. Thanks for the pointers!

eddierubeiz:
Oh - and here's more or less how I figured this out:

I went back and forth editing the inscriptions between these two URLS
```
	1) :3000/works/work_id
	2) :3000/concern/generic_works/work_id/edit
```

and simultaneously monitored the following 3 URLs by using watch and curl commands:
```
	3) :8080/rest/dev/xx/xx/xx/xx/work_id
	4) :8080/rest/dev/xx/xx/xx/xx/inscription_id
	5) :8983/solr/hydra-development/get?id=work_id
```

Once I figured that URL 1) was getting its data from Fedora, but 2) was getting its data from SOLR, I tweaked the indexing function inside the generic_work controller so as to make sure the inscription label was being updated prior to indexing. There may well be a better place to do this (perhaps in the indexer class for inscriptions) but this seemed to do the trick with minimal fuss.

jrochkind:
@eddierubeiz Makes sense, solr vs fedora issues are a frequent pain, and in my experience there's rarely a "better" place to fix em than whatever hack works, so I prob would have done what you did!

PR #1034, opened 2018-04-06
Breadcrumb cleanup

eddierubeiz:
Take a look when you have a minute, please! Suggestions welcome.


app/controllers/application_controller.rb
```diff
@@ -22,6 +22,26 @@ class ApplicationController < ActionController::Base
   # For APIs, you may want to use :null_session instead.
   protect_from_forgery with: :exception
 
+  #By default only show breadcrumbs to logged-in users.
+  def show_breadcrumbs
+    current_ability.current_user.logged_in?
+  end
+  helper_method :show_breadcrumbs
```
jrochkind: call it `show_breadcrumbs?`, question mark at the end is a nice convention for boolean methods, as seen in existing `logged_in?`

app/controllers/application_controller.rb
```diff
@@ -22,6 +22,26 @@ class ApplicationController < ActionController::Base
...
+
+  # This is used in generic_works_controller and collection_show_controller.
+  # It resolves issue https://github.com/sciencehistory/chf-sufia/issues/885
+  def special_breadcrumb_filter (the_breadcrumbs)
+    return if the_breadcrumbs == nil
+    # If the user's logged in, leave as is.
+    return if current_ability.current_user.logged_in?
+    # Otherwise, only show the "Back to Search Results" breadcrumb,
```
jrochkind: should be using the `show_breadcrumbs?` method you defined above, as `current_ability.current_user.logged_in?` Or do you mean to actually check logged in here, regardless of whether the controller has redefined `show_breadcrumbs?`? 

I don't entirely understand what this special filter is doing -- why isn't the `show_breadcrumbs?`method enough, why do you need to special filter even when you're showing breadcrumbs?

In general, this PR looks like a pretty sweet solution though, nice job!

app/controllers/application_controller.rb
eddierubeiz: Cool, will do!

app/views/_controls.html.erb
```diff
@@ -1,6 +1,10 @@
 <%# overrides https://github.com/projecthydra/sufia/blob/v7.3.0/app/views/_controls.html.erb %>
-<nav id="breadcrumbs" class="breadcrumbs-navbar-wrapper">
-  <%# custom CHF breadcrumbs up here instead of nav elements %>
-  <%= render_breadcrumbs builder: Sufia::BootstrapBreadcrumbsBuilder %>
-</nav><!-- /.navbar -->
-
+<% if show_breadcrumbs %>
+<%# show_breadcrumbs is a helper method
+  defined in application_controller and overridden
+  in a couple of subclasses.  %>
```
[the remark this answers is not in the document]
eddierubeiz: Got it!

app/controllers/application_controller.rb
jrochkind: Maybe just call the method `filter_breadcrumbs`?

Also standard ruby style is _no_ space in between method name and param paranthesis, in both method definitions and method calls.

app/controllers/application_controller.rb
```diff
@@ -22,6 +22,26 @@ class ApplicationController < ActionController::Base
...
+    # and suppress all other breadcrumbs. ( This method modifies
+    # the_breadcrumbs in place. )
+    back_str = I18n.t('sufia.bread_crumb.search_results')
```
jrochkind: comment "this method modifies the breadcrumbs in place" should go in comment above method definition, inside of inside.  Since it's important. 

I'm being extra detailed just as a learning exersize here, some of these are admittedly nitpicks!

eddierubeiz: Really appreciate your taking the time- very helpful!

app/controllers/application_controller.rb
```diff
@@ -22,6 +22,27 @@ class ApplicationController < ActionController::Base
...
+  #By default only show breadcrumbs to logged-in users.
+  def show_breadcrumbs?
+    current_ability.current_user.logged_in?
+  end
+  helper_method :show_breadcrumbs?
+
+  # This is used in generic_works_controller and collection_show_controller.
+  # It resolves issue https://github.com/sciencehistory/chf-sufia/issues/885
+
+  # Careful: this method modifies the_breadcrumbs in place.
```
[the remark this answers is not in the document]
eddierubeiz: NP!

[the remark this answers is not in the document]
jrochkind: PS: Command-`/` in sublime will add/remove '#' from beginnings of lines for you, super useful.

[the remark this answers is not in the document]
eddierubeiz: Awesome!

PR #1042, opened 2018-04-12
oai-pmh for dpla

jrochkind:


jrochkind:
Metadata may not be quite right yet, may need more tweaks after some more back and forth with PA Digital/DPLA. But merging to master for now, among other reasons so I can reindex with this code, which is needed.

PR #1046, opened 2018-04-17
Date modifications

eddierubeiz:
Please take a look when you have a minute - thanks!


jrochkind:
@eddierubeiz regarding our conversation:

We talked about whether the 'helper' object should take an _array_ of date models, or there should be an array of helper objects each with one date model (as in present code). 

While I suggested a single helper object that takes an array -- I'm not totally 100% on that. If it makes the code a lot more complicated for no clear immediate benefit, it may not be worth it. Your call as you develop. 

What I _am_ sure about is that the helper object(s) should take DateOfWork model(s) as input, rather than raw json hashes, and should not `< ActiveFedora::Base::TimeSpan` -- they can probably be pure ruby objects with no superclass other than Object. 

Make sense?

eddierubeiz:
Oh, it was good advice. It's already done in dev and working nicely. I'm wrestling with a test now but as soon as I'm done I'll push it to this branch. Only took an hour or two to refactor.

app/presenters/curation_concerns/generic_work_show_presenter.rb
```diff
@@ -58,6 +58,11 @@ def date_of_work_models
       end
     end
 
+    def display_dates
+      CHF::DatesOfWorkForDisplay.new(date_of_work_models).to_a
+    end
```
jrochkind: while I see what you were thinking and it is not at all unreasonable, I woudln't call this method `to_a`.  Is it returning strings or other objects? Let's turn it to a more descriptive name, "display_strings" or something. 

One gotcha with `to_a` is that some parts of ruby use it to _automatically_ turn the thing into an array when you don't expect them to, if it's not really an array-like object. 

Plus a more descriptive method is just good for letting the reader know what they're getting.

jrochkind: Actually maybe call the method inside DatesOfWorkForDisplay `display_dates` too, matching the presenter method?  Or call them both `date_display_strings`, or whatever.

eddierubeiz: Cool!

app/presenters/chf/dates_of_work_for_display.rb
```diff
@@ -0,0 +1,98 @@
...
+      if finish_string.blank?
+        date_string = start_string
+      else
+        # careful, this is an en dash:
+        date_string = [start_string, finish_string].compact.join(" – ")
+      end
+
+      if date_of_work.note.present?
+        date_string << " (#{date_of_work.note})"
+      end
```
jrochkind: Only noticed cause of the honeybadger error ha (no big deal! that's what staging is for). 

`<<` actually mutates the string. a non-mutating way to do it would be `date_string = date_string + whatever`, assuming you have a `date_string=` writer method. 

Mutating strings is generally considered undesirable unless you really want/need to do it for performance or whatever. 

Mutating strings passed in _by someone else_ is especially evil. 

Mutating strings passed in by someone else that are [frozen](http://blog.honeybadger.io/when-to-use-freeze-and-frozen-in-ruby/) will get you an error -- and thank whatever code froze it before passing it in, for doing so and preventing a much more mysterious error when one mutates someone elses string, heh.

PR #1047, opened 2018-04-19
WIP Batch edit changes

### LINKED ISSUE #925: Add Department metadata field to batch edit
Is there any way to do this easily? I have to make some 120+ edits and realized Department isn't an editable field in the batch edit workflow.

eddierubeiz:
Fixes #925 
This is only a first try, and probably contains bugs. Obviously a screwed-up batch edit attempt could ruin a lot of metadata, so it's worth testing this really carefully.
* I pulled method initialize_combined_fields from the superclass and modified it so it works with single-value fields.
* I'm also adding rights_holder and file_creator to this, as they seem to work largely the same way. Obviously we may not want them in the form, but I figured why not.

### DIFF
--- a/app/forms/batch_edit_form.rb
+++ b/app/forms/batch_edit_form.rb
@@ -95,6 +95,13 @@ class BatchEditForm < Sufia::Forms::BatchEditForm
     end
 
     def initialize_combined_fields
+      # This overrides the superclass,
+      # https://github.com/samvera/sufia/blob/v7.4.0/app/forms/sufia/forms/batch_edit_form.rb
+      # The goal here is to allow single-value fields to be edited via the form.
+      # To determine whether a field expects a single value or an array, we check the model's properties.
+      # If the field accepts an array, we process it as in the superclass. If it accepts a single value,
+      # then we process it slightly differently, storing the values in plain_attributes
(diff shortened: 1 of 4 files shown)


jrochkind:
I gotta admit that I don't understand this area of code at all, haven't looked at it before. 

That is a depressing amount of code to get this to work, but I have no trouble believing it's what's required in this stack. 

Did you copy and modify the `initialize_combined_fields` from the `super` implementation?  If so, the way I like to do it is:  a comment to that effect, explaining why you changed it, and providing a Github URL to the exact version in the dependency that you copied from.  (So if it changes in a future version and our copy breaks, it's easier to see that it changed and what changed).

eddierubeiz:
Correct -- I did copy and modify initialize_combined_fields from the superclass. Good policy re: adding a link to the code I copied. Will do!

jrochkind:
So I still don't really understand this code, but if it works, looks fine to me. 

Should we file a new github issue in 'backlog' for the bug you discovered? And let people know about it at weekly meeting, let them ("them") decide if it should be prioritized to fix or not. 

That's a really weird travis failure that I can't figure out why it's happening, or how it would be related to this new code. You can reproduce locally, on this branch but not master?

Issue #1049, opened 2018-04-25
Date displaying twice in record


catlu:
It's only entered once in the edit form. Seen here: https://digital.sciencehistory.org/works/9019s320q and here: https://digital.sciencehistory.org/works/ft848r50m.

archivistsarah:
I think this is happening only when adding a date to an existing work that didn't previously have the date field filled in, and not to dates added as part of ingest or editing an existing date entry. 

Appearing twice for this one (in which I added a date where there wasn't one before):
https://digital.sciencehistory.org/works/7h149p955

Appearing only once for this work (in which I changed "undated" to "1970s"):
https://digital.sciencehistory.org/works/1831cj92z

Works newly ingested & described this week are fine (ingested 4/26/2018): https://digital.sciencehistory.org/works/4b29b690d

jrochkind:
If I enter 1900, decade, I get "decade starting 1900s" which probably isn't quite English. 

Should probably be "decade starting 1900", or "1900s (decade)". The parenthetical qualifier seems a bit more readable to me for both century and decade, but no strong opinions. 

Also getting the duplicated dates when I edit a field too. Very mysterious! Wonder if this is a bug in _my_ Solr indexing stuff, hmm.

jrochkind:
We will want to get dates fixed before we send to PA Digital for DPLA, cause I'm using this stuff in the OAI-PMH field now!

jrochkind:
Okay, I guess ignore my issue if you can't reproduce!

eddierubeiz:
Hey @catlu and @archivistsarah  -- I think this is fixed in production. Please let me know if the bug comes back. I have reason to believe that it won't.

eddierubeiz:
Oh and thanks for testing last week, obviously!

PR #1053, opened 2018-05-02
Duplicate date bugfix

eddierubeiz:
This is a potential fix for #1049 .

Here's what I think is going on:

1) The generic work indexer is not deduplicating date objects before indexing them, as it does for some of the other fields it handles. This leads to duplicates in the SOLR field date_of_work_json_ssm .

2) The way the front-end display handled dates until a couple weeks ago masked this problem by deduplicating the date_of_work_json_ssm items at display time. ( I haven't investigated quite how yet.)

3) When [I changed the way front-end dates were displayed a couple weeks ago](https://github.com/sciencehistory/chf-sufia/commit/2556ae9cc3154025772a76c8d94ed4d4037c9d12), I used the data in date_of_work_json_ssm directly, thus exposing the problem described in 1).

I feel the best way to fix this problem is to modify the generic work indexer, which this PR does. My sense is that fixing this problem lower in the stack makes more sense, as duplicate dates in the date_of_work_json_ssm index could lead to more trouble down the line.

The data in Fedora doesn't appear to contain these duplicates, so, assuming we go this route, a reindex of the items edited since March 20th ( [when field date_of_work_json_ssm was introduced](https://github.com/sciencehistory/chf-sufia/commit/5db10bb75030c9c3ab0a419bfd01b5586375d4b5) ) should fix the problem. We could also run a full reindex, of course.

Quick note: I haven't yet made any further changes to date display in this PR, just to reduce confusion. (I did see your note about it in  issue 1049, @jrochkind , and of course I can easily make that change -- it's just I'd rather deal with this one first).

Also -- I added a regression test for this to the new work feature test.

To run *just* the feature test and then examine the solr field in dev, you might want to try a variation on the following:

```
./bin/rspec spec/features/new_work_feature_spec.rb
curl -vs http://localhost:8985/solr/hydra-test/select?indent=on  2>&1 | grep -5 2010
```

Hope all this makes sense -- let me know how yous feel.

### DIFF
--- a/app/indexers/chf/generic_work_indexer.rb
+++ b/app/indexers/chf/generic_work_indexer.rb
@@ -76,6 +76,9 @@ module CHF
         doc['visibility_ssi'] = object.visibility
 
         # index structured date of works, so we can get them at index time
+        # Not totally sure why we're getting duplicates here, the duplicates don't seem to make it to fedora,
+        # but we have dups as input here when indexing after an edit. We just de-dup here for now to avoid
+        # dups in index.
         doc['date_of_work_json_ssm'] = remove_duplicates('date_of_work').collect { |d| d.to_json(except: "id") }
 
         # Need all content-types for oai_dc serialization. This does force loading all members, was
--- a/app/indexers/chf/generic_work_indexer.rb
+++ b/app/indexers/chf/generic_work_indexer.rb
@@ -76,7 +76,7 @@ module CHF
         doc['visibility_ssi'] = object.visibility
 
         # index structured date of works, so we can get them at index time
-        doc['date_of_work_json_ssm'] = object.date_of_work.collect { |d| d.to_json(except: "id") }
+        doc['date_of_work_json_ssm'] = remove_duplicates('date_of_work').collect { |d| d.to_json(except: "id") }
 
         # Need all content-types for oai_dc serialization. This does force loading all members, was
         # that happening already, will it be a performance problem?
--- a/spec/features/new_work_feature_spec.rb
+++ b/spec/features/new_work_feature_spec.rb
@@ -91,6 +91,11 @@ RSpec.feature "Work form", js: true do
     expect(page).to have_text(artist_name)
     expect(page).to have_text(publisher_name)
     expect(page).to have_text("circa #{date}")
+
+    # Regression test for bug
+    # https://github.com/sciencehistory/chf-sufia/issues/1049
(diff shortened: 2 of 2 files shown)


jrochkind:
That makes sense, but I wonder how the duplicates end up in there in the first place -- we shouldn't really have duplicates in fedora _either_.  It sounds like maybe every time you press 'save', it duplicates all your dates?  Which is not good. 

We can/should definitely de-dup on indexing at least, and can deploy that right away, but maybe make another ticket for ourselves to figure out why the dates the indexer has as input are duplicated in the first place?  Are they duplicated in fedora, or just given to the indexer duplicated somehow, or what?  Are you interested in looking into that?

It also might not be unreasonable to say "it's the stack, it's always gonna have bugs, the best we can do is work around them, rather than spend all our time hunting them." That makes me sad, but I've had to do it before sometimes.

jrochkind:
PS: We can reindex _just works_ with `./bin/rake chf:reindex_works` on a `jobs` server. It takes 1-2 hours. Short enough that it's not worth trying to reindex just things that have changed since X date.

eddierubeiz:
* No, it appears the duplication is not happening in Fedora. It seems there's a bug between the work object and the indexer. I need to double-check that.

* Totally agree it's worth answering the question about where these duplicates are coming from. I'd like to be able to remove the dedupe function from the works indexer entirely. Yes, let's make another ticket -- and I'm happy to work on that ticket.

* As far as what currently gets duplicated, it appears to be the most recently added date only.

jrochkind:
Weird. It honestly might not be worth tracking down, it's just "the stack", de-duping in indexing may be sufficient. I have no idea, it's hard to know when to say when with this stack -- esp when we're running on an old version anyway, it may be different in newer ones, and the current one (we don't use yet) is about to be replaced by yet ANOTHER new one that will have possibly entirely different implementation. 

I'm going to make a commit to add a comment explaining it's mysterious to us why it needs de-duping, and then merge and deploy. 

Just curious -- did you confirm your regression test _failed_ without the fix, it really is a regression test?

And after I merge, do you want to do the reindex, so you know how to do it?

eddierubeiz:
* Sounds good.
* So glad you asked re: the test. I wrote the failing test first, because they drummed TDD into our heads in Portland, and I was like - why the hell not.
* Yeah, happy to reindex. I've done it in dev a couple times. I would like to try it on staging first, assuming we can push this fix to staging, just to double-check that it fixes a few test items.

Issue #806, opened 2017-08-28
Upgrade to ruby 2.4.1


jrochkind:
Deployment machines are currently on 2.3.1. 

MRI 2.4 has some non-trivial performance improvements, and some relatively minor new stdlib API that it would still be nice to be able to use. 

I don't anticipate any backwards breaking changes to any of our code, it's more about how painful/time-consuming it will be to make the change on our deployment machines.

jrochkind:
probably upgrade passenger while we're at it. https://blog.phusion.nl/2018/01/29/passenger-5-2-0/

jrochkind:
I actually found a bug in 2.5, so probably wait until 2.5.1.

sanfordd:
Testing worked well, we'll migrate staging this week or early next week and do production next week.

PR #1055, opened 2018-05-10
update devise to a version that works under ruby 2.5.1

jrochkind:
https://github.com/plataformatec/devise/issues/4719

After this update, all our tests pass on ruby 2.5.1. 

Ref #806


eddierubeiz:
Merging after cursory review.

PR #1056, opened 2018-05-14
WIP have solr_wrapper use mirror_url

jrochkind:
Travis can't pull down solr, so all our tests fail. 

We think this is because apache.org is blocking/rate-limiting travis IPs. 

We think that using a new solr_wrapper `mirror_url` feature may fix. This is a test to see if it will. 

New solr_wrapper mirror_url feature is not yet in a released solr_wrapper, so this branch uses solr_wrapper from git. We'd really like to get a solr_wrapper release with the feature, if using it resolves.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -142,7 +142,10 @@ group :test do
 end
 
 group :development, :test do
-  gem 'solr_wrapper', '>= 0.3'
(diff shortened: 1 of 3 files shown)


jrochkind:
Problem not currently occurring, plus some reason to believe this won't actually fix it anyway, and another PR to solr_wrapper is required. https://github.com/cbeer/solr_wrapper/issues/116

PR #1058, opened 2018-05-14
WIP - fix broken timeout for batch edit form feature test.

eddierubeiz:
This is an attempt to fix a problematic front-end test for the batch edit form. The test changes various fields, then clicks "Save", and waits for the words "Changes Saved" to appear on the screen before proceeding.
This test passes on my laptop but fails when run by the Travis machines. So far, increasing the amount of time it waits hasn't yielded any results.


jrochkind:
we are stumped, can't figure it out, are giving up for now.

Issue #6, opened 2015-04-15
Create admin page to monitor fixity logs and checksums


hackmastera:

hackmastera:
relevant, in-progress: https://github.com/projecthydra/sufia/pull/1181/files

jrochkind:
would want some specs on what we'd want here. Just failing fixity checks? (which will usually be zero).  Interactive ability to look up fixity checks on things to confirm when they last happened etc? Other? 

Should think in terms of our staff use cases, what we need/want/will use this information for.

MDiMeo:
This first came up years ago when I was completing my digital curation certificate. As part of that class, we were testing various preservation software packages and these dashboards often had a way to check fixity. I remember taking screenshots, but we weren't sharing stuff like that in GitHub tickets back then. Let me do some digging...

But in the meantime, my use case is that I would like a place where I could glance at recent fixity checks and clearly see that they had all passed. I would imagine that a report would be generated automatically and that this would just be a place where I could glance and make sure that fixity checks were running and passing. If something failed, it would be flagged up in a really obvious way in the UI and, in an ideal world, I would have also gotten an email (which would prompt me to go on here.) I would want to see the object number  and when it failed so that I could talk to @sanfordd about locating a back-up.

jrochkind:
There isn't really a "report" generated, there's just a database of all the recent fixity checks (we can configure how recent we keep, but it doesn't keep them all forever to avoid filling up the db).  and their statuses.  

I'm thinking the 'dashboard' kind of is the report, and should probably a) Definitely tell you if there are any failures, with detials, b) give some kind of summary information about how many fixity checks have been run recently, to keep you confident that fixity checks are happening. 

You can already navigate to an individual file 'info' page to find out when the most recent fixity check was run on it, and it's status.

jrochkind:
hmm, maybe also ideally information on the files that have currently gone longest without a fixity check... but that's pretty hard to figure out from the sufia data model, alas.

MDiMeo:
Yes, this all sounds like it's in line with my thinking. I remember that one of the systems I tried had allowed me to enter date limits and it generated a rather unwieldy spreadsheet (for example, all fixity checks run over the past month.) I'm not sure I can think of a use case for this though.

It would be nice to know which has gone the longest without a fixity check. Or some way to generate a query to see what hasn't been checked for [xx] months. But I think of this as a bonus feature. I'm mostly concerned with points A and B in your comment, above.

[...]

PR #1063, opened 2018-05-21
WIP Fixity dashboard

eddierubeiz:
Still needs a test for the controller.


eddierubeiz:
Fixes #6

lib/tasks/clean_up_orphaned_fixity_records.rake
```diff
@@ -0,0 +1,61 @@
...
+# still have on hand. If it turns out they are not, delete the "orphaned" fixity check record.
+
+namespace :chf do
+  desc "run fixity checks with logs and notification on failure"
+  task :clean_up_orphaned_fixity_checks => :environment do
+    @conn = Blacklight.default_index.connection
+    x = 0
+    n = 100
+    loop do
+      batch = ChecksumAuditLog.select(:file_set_id).distinct[x..x+n-1]
```
jrochkind: Not actually sure what this is doing, but I don't think it's something good.  You're trying to fetch all the rows, but in batches to avoid having them all in memory at once?  Good plan. 

The right way to loop through all rows in an rdbms with ActiveRecord would be using the `limit` and `offset` methods. http://guides.rubyonrails.org/active_record_querying.html#limit-and-offset

However, I think we can do one better, and just use ActiveRecord's find_in_batches to do it for us. 

I am not sure if you are trying to loop through _all_ the rows, or just the ones representing most recent checks?

```ruby
ChecksumAuditLog.find_in_batches(batch_size: 100) do |array_of_one_hundred|
# or perhaps
ChecksumAuditLog.latest_checks.find_in_batches(batch_size: 100) do |array_of_one_hundred|
```
Alternately to find_in_batches, if it is more convenient to yield each row individually, but still fetch in batches:

```ruby
ChecksumAuditLog.find_each(batch_size: 100) do |checksum_audit_log_record|
ChecksumAuditLog.latest_checks.find_each(batch_size: 100) do |checksum_audit_log_record|
```
http://api.rubyonrails.org/classes/ActiveRecord/Batches.html#method-i-find_each

The default batch size of 1000 _may_ be just fine in either case, but 100 is fine too. 

Having trouble understanding the rest of the code, how about you start refactoring to use one of these approaches, and then I can re-review?

(In general, if you find yourself using `loop do` in ruby, there's _probably_ a better way).

jrochkind: (Oh, I see, it's 100 becuase you may be checking them all in solr at once? ok, 100 is fine!)

lib/tasks/clean_up_orphaned_fixity_records.rake
```diff
@@ -0,0 +1,56 @@
...
+#
+# What this job does:
+# Loop through all fixity records and verify that they are associated with a fileset that we
+# still have on hand. If it turns out they are not, delete the "orphaned" fixity check record.
+
+namespace :chf do
+  desc "run fixity checks with logs and notification on failure"
+  task :clean_up_orphaned_fixity_checks => :environment do
+    @conn = Blacklight.default_index.connection
+    ChecksumAuditLog.find_in_batches(batch_size: 100 ) do | batch |
```
eddierubeiz: Thanks for the helpful advice. Refactored to use find_in_batches. Gee, SOLR is really really fast.

app/views/sufia/admin/_sidebar.html.erb
```diff
@@ -0,0 +1,40 @@
+<ul class="admin-sidebar nav nav-pills nav-stacked">
```
jrochkind: This is overriding the entire .html from sufia just to add the link to the new page?  That is fine (and probably the only way to do it), but add a comment to that effect, as well as a github URL to the exact version of the original sufia file you copied from, please and thank you!

lib/tasks/clean_up_orphaned_fixity_records.rake
```diff
@@ -0,0 +1,56 @@
...
+    []
+  else
+    ids_returned = response['response']['docs'].collect{ |x| x['id'] }
+    list_of_file_set_ids - ids_returned
+  end
+end
+
+def item_exists_in_fedora?(file_set_id)
+  begin
+    FileSet.exists?(:id => file_set_id)
```
jrochkind: Are we positive this really checks Fedora and not Solr? I never know which to expect.

app/views/sufia/admin/_sidebar.html.erb
eddierubeiz: Correct! Will do, no problem at all.

lib/tasks/clean_up_orphaned_fixity_records.rake
eddierubeiz: I'm pretty confident but not 120% sure. I will find a way to triple-check before we proceed.

eddierubeiz: Thanks for the gentle hint. This is not hitting Fedora. Stay tuned...

PR #1066, opened 2018-05-23
Footer changes (copyright and link to privacy policy)

eddierubeiz:


app/views/shared/_footer.html.erb
```diff
@@ -36,4 +36,10 @@
       <div class="footersocialicon"><a href="https://www.instagram.com/<%= Rails.application.config.instagram_acct %>" target="_blank"><i class="fa fa-instagram" aria-hidden="true"></i></a></div>
     </div>
   </div>
+  <div class="footerrow">
+    <div class="footerprivacycopyright">
+      &copy; <%= Time.current.year %>  Science History Institute |
+      <a href="https://www.sciencehistory.org/privacy-policy/" target="_blank" >Privacy Policy</a>
```
jrochkind: Do you think we need a target _blank on this? The main site doesn't have one. 

In general, apparently scihist style is target blank on any "external" link, but this isn't really an external link, it's all sciencehistory.org -- or might they get confused without the digi collections navbar once they get there?

In general, I'm big on avoiding target blank unless really needed. Without target blank, someone who knows how can choose to open in a different window (and there's always the back button). With it, you get an extra tab/window whether you like it or not, which I think can be esp annoying on mobile. 

Thoughts?

jrochkind: @RonBrashear @catlu or Hillary may have a preference.

eddierubeiz: If we consider the two domains part of the same site, let's remove the target _blank. Two seconds...

RonBrashear: My preference would be to not have target_blank. The concept should be to make it seem like it's part of the SHI site.

eddierubeiz:
OK - removed the target = _blank. @jrochkind, if no other feedback, feel free to merge.

PR #1076, opened 2018-05-30
browse-everything support

jrochkind:
With extensive refactoring of actor/jobs stuff for ingest that turned out kind of neccesary to make it work right. :(


jrochkind:
This majorly refactors ingest behavior, but I haven't written any tests. :( I just got overwhelmed, there are _so_ many moving parts, and testing file ingest is hard (and slow) to begin with. It's possible Scholarsphere has some we could copy or crib from.  It's possible that should be a pre-req for merging this, not sure.

jrochkind:
There is half-finished support in sufia for actually giving you some status/progress info on ingests as the files are slowly added. It, I believe, remains half-finished and not doing too much in Hyrax, so we wouldn't have gotten it by upgrading to hyrax. 

This would be a nice thing to have, it is not included in this PR at present, we can talk about if we need to/want to prioritize providing such a UI, either based on existing stack code or not.

jrochkind:
Docs of what/why/how, with diagrams of architecture: https://wordpress.com/post/bibwild.wordpress.com/7785

Issue #1084, opened 2018-06-06
Use postgres in dev and test


jrochkind:

jrochkind:
Closed by #1086

PR #1086, opened 2018-06-07
Use postgres in dev and test, instead of sqlite3

jrochkind:
Be more consistent with prod, and sqlite3 has problems with lots of concurrent access sometimes, which sufia ingest can do.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,10 @@ cache:
     - $HOME/.phantomjs
 before_script:
   - jdk_switcher use oraclejdk8
+  - psql -c 'create database scihist_sufia_test;' -U postgres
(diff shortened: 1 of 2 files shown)


jrochkind:
Closes #1084

jrochkind:
@eddierubeiz you wanna try it out on this branch before we merge, to make sure it works as smoothly for you as it is for me, and we're not missing anything? No rush.

eddierubeiz:
I'm already doing just that ...

eddierubeiz:
Yup - works great!

jrochkind:
ok, if you feel ready to deal with it, merge whenever!

PR #1098, opened 2018-06-13
Recent Items

### LINKED ISSUE #538: implement 'recent items' section of front page

eddierubeiz:
Here's a very simple proof of concept that we can use to have a more grounded discussion about what we want. I decided not to spend too much time worrying about the design before we have more specifics, so it still looks sloppy.

For now, I'm taking 15 recent works, dumping them into a bag, shuffling the bag, and taking 6 out of them to show on the front page. I reshuffle the bag every minute and pick 6 new items-- but out of the same bag of 15 recent works. All three of these parameters are easy to change, but I had to start somewhere.

I'm calling the new section "Featured items". The section below it (which formerly didn't have a title, now also gets a title: "Collections".

Fixes #538 .

### DIFF
--- a/app/controllers/sufia/homepage_controller_override.rb
+++ b/app/controllers/sufia/homepage_controller_override.rb
@@ -22,7 +22,7 @@ Sufia::HomepageController.class_eval do
   helper_method :public_works_count
 
   def recent_items
-    how_many_works_to_show = 5
+    how_many_works_to_show = 6
     how_often_to_change = 60 * 10 # ten minutes in seconds
 
     # @@arbitrary_number is a slowly incrementing integer that changes at most every
--- a/app/controllers/sufia/homepage_controller_override.rb
+++ b/app/controllers/sufia/homepage_controller_override.rb
@@ -22,7 +22,7 @@ Sufia::HomepageController.class_eval do
   helper_method :public_works_count
 
   def recent_items
-    how_many_works_to_show = 8
+    how_many_works_to_show = 6
     how_often_to_change = 60 * 10 # ten minutes in seconds
 
     # @@arbitrary_number is a slowly incrementing integer that changes at most every
--- a/app/controllers/sufia/homepage_controller_override.rb
+++ b/app/controllers/sufia/homepage_controller_override.rb
@@ -22,11 +22,11 @@ Sufia::HomepageController.class_eval do
   helper_method :public_works_count
 
   def recent_items
-    how_many_works_to_show = 12
+    how_many_works_to_show = 8
     how_often_to_change = 60 * 10 # ten minutes in seconds
 
     # @@arbitrary_number is a slowly incrementing integer that changes at most every
-    # how_often_to_change minutes. When it does change, we a new bag of recent works
+    # how_often_to_change minutes. When it does change, we fetch a new bag of recent works
     # from SOLR and reshuffle the bag even if there haven't been any new works added.
     new_arbitrary_number = Time.now.to_i / how_often_to_change
     if (!defined? @@arbitrary_number) || (@@arbitrary_number != new_arbitrary_number)
--- a/app/controllers/sufia/homepage_controller_override.rb
+++ b/app/controllers/sufia/homepage_controller_override.rb
@@ -46,7 +46,7 @@ Sufia::HomepageController.class_eval do
     rescue
       return view_context.link_to("Image not in this repository", "#")
     end
-    link_to(solr_doc.title.first, curation_concerns_generic_work_path(solr_doc.id))
+    view_context.link_to(solr_doc.title.first, curation_concerns_generic_work_path(solr_doc.id))
   end
   helper_method :featured_collection_image_link
 
--- a/app/controllers/sufia/homepage_controller_override.rb
+++ b/app/controllers/sufia/homepage_controller_override.rb
@@ -22,7 +22,7 @@ Sufia::HomepageController.class_eval do
   helper_method :public_works_count
 
   def recent_items
-    how_many_works_to_show = 5
+    how_many_works_to_show = 12
     how_often_to_change = 60 * 10 # ten minutes in seconds
 
     # @@arbitrary_number is a slowly incrementing integer that changes at most every
--- a/app/controllers/sufia/homepage_controller_override.rb
+++ b/app/controllers/sufia/homepage_controller_override.rb
@@ -22,7 +22,7 @@ Sufia::HomepageController.class_eval do
   helper_method :public_works_count
 
   def recent_items
-    how_many_works_to_show = 6
+    how_many_works_to_show = 5
     how_often_to_change = 60 * 10 # ten minutes in seconds
 
     # @@arbitrary_number is a slowly incrementing integer that changes at most every
--- a/app/controllers/sufia/homepage_controller_override.rb
+++ b/app/controllers/sufia/homepage_controller_override.rb
@@ -13,7 +13,6 @@ Sufia::HomepageController.class_eval do
     SearchBuilder::HomePage
   end
 
-
   protected
 
   def public_works_count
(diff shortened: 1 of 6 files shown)


app/views/sufia/homepage/_recent_items.html.erb
```diff
@@ -0,0 +1,30 @@
...
+works_to_pick_from = GenericWork.search_with_conditions( conditions, opts)
+
+# Now, pick a few of these out of the bag at random to show.
+# Shuffle the bag again every few minutes.
+how_often_to_change=60 #seconds
+srand Time.now.to_i/how_often_to_change
+how_many_works_to_show = 6
+works_to_show = works_to_pick_from.sort_by{rand}[0..how_many_works_to_show-1]
+%>
+
```
jrochkind: this should not be inline in the view. It should be in the controller or a helper method.

jrochkind: I am not sure what your "how often to change" implementation is doing, or if it's working.   What do you think it's doing?

jrochkind: if we're going to be doing the fetch every time anyway, might as well just have it change on every page load. The only point I see to having it say 'stable' for eg 60 seconds is to avoid the performance hit of fetching em on every view. Is it bad UX if it changes on every page load?

app/views/sufia/homepage/_recent_items.html.erb
```diff
@@ -0,0 +1,30 @@
...
+        <%= link_to main_app.curation_concerns_generic_work_path(work['id']) do %>
+          <%= image_tag  work['thumbnail_path_ss'], class: "recent-items-image", width: "100%"%>
+          <div class="recent-item-title-wrapper">
+          <p class="recent-item-title"><%= work['title_tesim'].first %></p>
+          </div>
+        <% end # link %>
+    </div>
+    <% end # each %>
+</div>
+
```
jrochkind: Let's call this "Recently Added Items" (Or "Some of our New Items" or something), instead of "Featured Items". They aren't really intentionally "featured".

app/views/sufia/homepage/_recent_items.html.erb
eddierubeiz: Will move to the controller, of course! Just wanted to have everything in one place as I mess with the design a bit; I'm already changing thing around.
I do like having things change a bit more slowly than on every page view -- ideally I'd like it to change every 10 minutes or so but I wanted to make sure it was working and it's easier to test if tit changes every minute.
The use case for having things change a bit more slowly is: "I found something cool serendipitously on the home page, and I want it to be there in a couple more minutes once I switch e.g. from my phone to my desktop."

app/controllers/sufia/homepage_controller_override.rb
```diff
@@ -13,10 +14,42 @@ def search_builder_class
...
+    how_many_works_to_show = 5
+    how_often_to_change = 60 * 10 # ten minutes in seconds
+    how_many_works_in_bag = 15
+
+    #First, put a bunch of eligible works into a bag.
+    # Note that the call to SOLR is memoized.
+    conditions =  {'read_access_group_ssim'=>'public'}
+    sort_by = ["system_modified_dtsi desc"]
+    opts = {:rows=>how_many_works_in_bag, :sort=>sort_by}
+    @works_to_pick_from ||= GenericWork.search_with_conditions( conditions, opts)
```
jrochkind: since this iVar is on the whole controller, call it something that makes sense out of context. perhaps `@recent_items_to_pick_from`, or even just `@recent_items`. 

But also, instance variables on a controller are only good for the duration of a request, since a new Controller instance gets created on every request. So this won't actually cache the value between requests, if that's what you were trying to do. 

You need to store it in some kind of "global" state, such as a class variable or rails `class_attribute`.

eddierubeiz:
Hey @jrochkind  -- can you clarify what you were thinking in  https://github.com/sciencehistory/chf-sufia/pull/1098#issuecomment-398479050  ? What would you change about the screenshot? Were you thinking of having more items on the second line or zero horizontal space between the first and second line or something else?

jrochkind:
yeah, just less horizontal space and maybe left-justifying. 

here's just a hacky graphical edit. 

screen shot 2018-06-19 at 1 21 16 pm

jrochkind:
I would try to do it by taking the margin off the individual cells, and adding padding to the wrapper instead. And then some trickery to make sure it's left-justified when it wraps, perhaps requiring an additional level of div wrapper.

eddierubeiz:
Got it! Stay tuned.

app/helpers/recent_items_helper.rb
```diff
@@ -0,0 +1,89 @@
+# A random selection of recently modified works to display as thumbnails
```
jrochkind: Nice job using actual object-oriented design (the RecentItems class/object), instead of just a bunch of flat helper methods, like Rails tries to get you to. 

There's really no reason this _needs_ to be in ./app/helpers -- the _standard_ thing to be in ./app/helpers would be that formentioned flat list of Rails 'helper' methods (which get mixed into the 'view context').  

But I don't see any reason it would cause a problem, and don't know any reason _not_ to put it here, so that's fine.

app/helpers/recent_items_helper.rb
```diff
@@ -0,0 +1,89 @@
...
+    end
+
+    def var_missing?(symbol)
+      begin
+        tmp = self.class.class_variable_get(symbol)
+        defined?(tmp) != "local-variable"
+      rescue
+        true
+      end
+    end
```
jrochkind: You don't really need this. 

a) if you get to that part of the code`defined?(tmp)` is actually _always_ gonna be `local-variable`, there's no purpose for the check. 
b) You don't need to get class variables as `self.class.class_variable_get`, unless you actually need some indirection/reflection. You can just ask for `@@class_variable`. 
c) I _thought_ you could just ask for `@@class_variable` and if it didn't already exist it would just return nil. I was surprised to try it in irb and see that indeed (unlike @instance_vars) that isn't the case!  Which probably motivated all this code...

HOWEVER. I think the right way to deal with it is... just to initialize all the class variables you want at the top of the class. 

```
class RecentItems
  # not positive what the num variable stores, can you rename this more descriptively too? Is it actually "last_refresh_timestamp"? 
  @@num = nil  
  @@bag = nil
```

If you just declare them like that, ruby will never complain about you accessing them, and you won't need all that self.class.class_variable_get/rescue business. Or the `_missing?` methods. 

_Really_ I'm not sure if you need any cover methods at all at that point. Do you need:

```ruby
def bag
  @@bag
end
```

Or can your code just access `@@bag` directly?  I think probably directly is fine for this little self-contained thing, although an argument could be made for the cover methods. BUT in ruby, when we do getters/setters, the convention is simply `def num` instead of `def get_num`, and `def num=(v)` instead of `def set_num`. 

All of this stuff is so self-contained in this case that we really _could_ just deploy it, it's not going to "infect" anything else. So if you're really sick of working on this, we could just do that, or you could take another pass and simplify the code a bit (remove a bunch of those methods!).  Your call.

app/helpers/recent_items_helper.rb
```diff
@@ -0,0 +1,89 @@
...
+      trigger_reshuffle_if_needed()
+      # First, put a bunch of eligible works into a bag.
+      works_to_pick_from = bag_of_recent_items()
+      # Now, pick a few of these out of the bag at random to show.
+      # Reshuffle the bag every now and then.
+      srand get_num()
+      works_to_pick_from.sort_by{rand}[0... @how_many_works_to_show]
+    end
+
+    private
```
jrochkind: nice job marking things `private` when not meant to be external API

app/helpers/recent_items_helper.rb
jrochkind: Or _I_ could take a pass at simplifying it to show you how I'd do it, as a compare and contrast. Whatever you like!

eddierubeiz: No, I definitely should do it! I'll refactor based on your suggestions, which are great Stay tuned - I'll need a break from this collections bug soon, and this will be fun and easy by comparison! :)
Thanks for the great feedback.

app/helpers/recent_items_helper.rb
```diff
@@ -0,0 +1,49 @@
...
+    def initialize()
+      @how_many_works_to_show = 6
+      #@how_often_to_change = 60 * 10 # ten minutes
+      @how_often_to_change = 5
+      @how_many_works_in_bag = 15
+    end
+
+    def recent_items()
+      @@bag = nil if time_for_a_new_bag?
+      srand @@last_refresh
```
jrochkind: I'm sorry I'm just noticing this, but I think srand will seed the ruby pseudo-random number generator _program wide_, you've changed how random numbers are generated everywhere. 

Which also might not result in the repeatability you want, if other parts of the program use the random number generator in between your calls to it, which they very well might. 

I wonder if there's a way to do this with just local effects.

eddierubeiz:
I wonder.... stay tuned...

app/helpers/recent_items_helper.rb
jrochkind: Looking at ruby docs, I think maybe we can do this:

```
generator = Random.new(seed)
generator.rand
```

To use a local one with our custom seed, instead of insisting on our custom seed program-wide. 

I guess this `generator` would _also_ have to be stored in a class variable, so repeated calls to it are to the same generator and get the sequence. 

I think this all shows how doing "clever" things incures a cost, more possibility for bugs and more time spent making it robust and reliable. I wonder if we want to do this clever rand thing at all, but I think let's stick with it for now to hopefully finally get done with this feature!

eddierubeiz:
OK - I'm trying the solution in https://stackoverflow.com/questions/23047911/how-to-shuffle-an-array-the-same-way-every-time and this seems to work nicely.
Point well taken! If I hadn't learned so much in the process it would probably not have been worth it.

jrochkind:
That solution looks pretty good if it works, great!

PR #1103, opened 2018-06-14
New server ip

sanfordd:
New IP for the solr box with the Ruby 2.5.1 migration.


sanfordd:
Opening a new one one without all this extra stuff.

Issue #1069, opened 2018-05-29
Report to identify records marked public w/o rights statement


HKativa:
Reminder: DPLA requires a rights statement; depending on when this ticket is addressed, we may have to omit such records from our OAI-PMH feed, and/or let PA Digital omit them from what it sends to DPLA.

eddierubeiz:
Good news! Based on a cursory check, it appears that all public works have rights statements at the moment. (I double-checked with Cat, and she confirmed that the one problematic item I was able to find in Staging had in fact been flagged by the DPLA folks at Temple and corrected in the mean time in Production.)

Should we want to run this cursory check again periodically, I've added the recipe to the wiki at: https://chemheritage.atlassian.net/wiki/spaces/HDCSD/pages/15368194/Application+administration .

@HKativa , @catlu , @jrochkind , if we want to turn this code into a permanently available page in the admin section, I can of course do that, but it seems like overkill at the moment. Your thoughts are welcome.

HKativa:
@eddierubeiz Thanks for the update. Agreed that turning code into page on admin section isn't necessary at this point and the documentation you created should suffice.

eddierubeiz:
See above for a possible solution to the problem. This would automatically prevent users from publishing works w/o a rights statement.

PR #1106, opened 2018-06-18
WIP Prevent items without right statements from being marked public.

eddierubeiz:
Prevents you from selecting the radio button marked "Public" without first picking a rights statement.  For now, this takes the form of a few lines of client-side code that are executed when the generic work form is loaded; I suspect there's a better way to do this but I'm starting with the obvious.
Related to #1069 .


jrochkind:
not going to fix this for now, and not going to do it like this.

PR #1115, opened 2018-06-20
Locking out users

eddierubeiz:
Fixes https://github.com/sciencehistory/chf-sufia/issues/1107 . We won't add tests for this one (see below.)

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -38,14 +38,18 @@ class User < ApplicationRecord
   # A devise method, we're taking advantage of it for a cheesy way to lock out
   # admin users based on CHF::Env variable.
   def active_for_authentication?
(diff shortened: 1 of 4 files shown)


jrochkind:
Awesome. You may already be on this, but also add a rake task we can use to disable someone (and maybe another one to enable someone), seem cool?

jrochkind:
(This is one _I_ wouldn't do tests for honestly, heh. It's a devise feature, we assume our dependencies work. But it's totally up to you, test or not. One of the downsides of tests with this stack is taht they are _slow_, so every test we add (esp an integration test like this may need to be), slows down the end-to-end time of our total suite, which is annoying. But I leave it up to you, if you think a test is needed!)

jrochkind:
looks great, thanks @eddierubeiz !

PR #1116, opened 2018-06-22
Facet bug fix experiment

eddierubeiz:
Took a little time to investigate https://app.honeybadger.io/projects/53196/faults/38353008 and this is a quick possible fix for that bug.
On the one hand, this is a very small bug, and I understand we don't want to make it a habit of attempting to fix bugs like this. On the other, it looks like it's easily remedied by adding a bit of error-checking in one of our own methods.

### DIFF
--- a/app/controllers/collections_show_controller.rb
+++ b/app/controllers/collections_show_controller.rb
@@ -24,6 +24,7 @@ class CollectionsShowController < CatalogController
   # displays values and pagination links for a single facet field
   def facet
     @facet = blacklight_config.facet_fields[params[:facet_id]]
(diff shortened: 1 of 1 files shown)


jrochkind:
Hey @eddierubeiz , I was ignoring this cause of "experiment" in it's name, but it looks like it's actually good to go, and you recommend merging?

eddierubeiz:
Sure, why not - go ahead.

jrochkind:
don't remember enough about what this is, aren't completely sure it won't break, just closing it. It was never prioritizied, just an eddie experiment.

PR #1120, opened 2018-06-26
Fix bug with batch ingest with formats selected

### LINKED ISSUE #1114: Unable to batch ingest using local files option
Scenario: You are an admin. Under "Works," you select "Batch Create." Under "Add Local Files," you click "Add Folder" and select a folder on the M drive. Under "Descriptions," you enter metadata for "Identifier," "Creator," "Date of Work," "Genre String," "Extent," "Subject," "Rights," and "File creator." You set "Visibility" to "Your Institution." Under "Share," you share editing privileges with another user. After clicking save, you are redirected to the homepage for your works, advised that files are processing in background and you may need to refresh page to see them. However, files never show up. No user notifications that files were imported.

Bug manifested after merging new ingest function for S3. Ingests following this process failed in both production and staging.

jrochkind:
Closes #1114

Our code was wrapping an extra array where it shouldn't have, not sure why, something lost in translation of copying and customizing sufia code.

In testing period, nobody tested batch upload of _local_ files with 'formats' selected, which is the only time it would occur.

### DIFF
--- a/app/jobs/batch_create_job.rb
+++ b/app/jobs/batch_create_job.rb
@@ -37,10 +37,9 @@ class BatchCreateJob < ActiveJob::Base
     def create_uploaded_files(titles, resource_types, attributes, uploaded_files = [])
       uploaded_files.each do |upload_id|
         title = [titles[upload_id]] if titles[upload_id]
-        resource_type = [resource_types[upload_id]] if resource_types[upload_id]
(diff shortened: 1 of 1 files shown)


jrochkind:
A line and a half of bugfix, ~4 hours to figure it out. :)

Issue #1105, opened 2018-06-18
Attaching child works to parent work generates software error on save


HKativa:
Scenario: You are an admin. From your "My works" page, you select a work to edit. Under the "Relationships" tab, you relate several child works to this parent work using filenames in the "search for a work" feature. Upon saving the record, you are redirected to the "Software Error (500)" page. However, when you return to the "My works" page and reopen the parent work, all child works are now related and show up in the record.   

Specific example: Adding several child works to work ID c247ds983 generated Error ID: 3f0c467a-0533-4bcf-aaa8-3290db33d0e8 in Honeybadger.

HKativa:
Note: I encountered this error with several records (ID c247ds983; ID dn39x222c; ID ks65hd14w). In one instance, ID c247ds983, the files for the child works show up in the parent work, but the "part of" relationship does not display in the records for the individual child works. The "part of" relationship does display on the child work records for ID dn39x222c and ID ks65hd14w).

eddierubeiz:
Just adding [a link to the HoneyBadger error](https://app.honeybadger.io/projects/53196/faults/37830755/a456abb4-7315-11e8-8225-5c6342de622c#notice-summary:).

HKativa:
Possibly related issue - see #1112: after relating child works to parent ID 8623hz59c generated software error on save (see above), child works show up in parent record, but relationship is not present in child records. Child record IDs:
8049g5923
5x21tg194
9880vr89z
f1881m93b
gm80hw154
bk128b84c
h415pb36g

jrochkind:
In my dev copy, I am strangely having trouble getting the "search for a work"-to-add-childworks feature to work at all. nothing shows up in my autocomplete, even though I have other works. 

Anyone have any idea what's up with that?

jrochkind:
okay got it working with some new works i created in my dev copy. 

succesfully reproduced in dev.

eddierubeiz:
Just tested the above and ""search for a work"-to-add-childworks" appears to work fine for me.

jrochkind:
This is possibly a known bug in ActiveFedora, that maybe has been fixed in later versions we aren't using yet. https://github.com/samvera/hyrax/issues/2108

Still trying to figure out for sure if that's what it is, and what the fix was inside ActiveFedora. 

What's mysterious to me is how we never encountered this before. I wonder if there is a different UI element @catlu et al had been using to add child works, and this one was always broken? I don't think that's it, but it's mysterious.

[...]

PR #1121, opened 2018-06-27
upgrade active_fedora to 11.5.2

jrochkind:
From 11.1.6

To upgrade activefedora we also had to upgrade other dependencies, so as to be compatible.

This is to get a version of active-fedora which fixes this bug:
https://github.com/samvera/hyrax/issues/2108

Which seems to be responsible for this bad behavior in our app:
https://github.com/sciencehistory/chf-sufia/issues/1105

It's a mystery why we didn't run into this active-fedora bug until now, but this does seem to fix it.

However, upgrading active-fedora required upgrading faraday, which made some automated tests fail around our "OPAC Service" scraping functionality. There was a comment note from @hackmastera saying it would, although it's mysterious why.

I believe we do not actually use that functionality anymore, and can probably just remove those tests and/or all the code behind that feature. Alternately we could make it work again.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -11,7 +11,7 @@ gem 'blacklight_range_limit', ">= 6.1.2"
 # Update: slackarchive permalink doesn't work after some max. search for 'rsolr' and look
 # at April 7 results for hints of the conversation? Unfortunately you can't get context.
 gem 'rsolr', '~> 1.0'
-gem 'active-fedora', '~>11.1.6'
+gem 'active-fedora', '~>11.5.2'
 gem 'curation_concerns', '~>1.7.7'
 
 # pull in fix to "add another" label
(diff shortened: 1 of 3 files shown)


jrochkind:
Closes #1105

jrochkind:
I think the test failures are acutally just _test infrastructure failures_, around changing version of faraday requiring you to refresh the webmock "playbackable transactions". They don't mean the feature is actually broken. 

On the other hand, I believe feature _was_ actually broken, although not caught by tests (And also unused). 

And fixing the Webmock stuff is a confusing pain, especially with me being 'off campus', not sure if I actually have network access to whatever the tests around "OPAC service" were accessing...  hmm, looks like `https://sandbox.iii.com/iii/sierra-api/v2/token` so maybe not a problem.. but it looks like it ALSO needs an API key that I have no idea where it comes from. 

This was all Anna's stuff that i know little about, and I believe the feature has been broken, with nobody really minding, for quite some time. So I'd be inclined to just disable the broken tests.

PR #1124, opened 2018-06-29
New drop-down value for Creator field

### LINKED ISSUE #1122: New drop-down value for Creator field
Add "Manner of" to the drop-down for "Creator" field, after "Interviewer" and before "Manufacturer"

eddierubeiz:
This appears to be working nicely now. Due to a Fedora bug we can't use a Getty URL for the "manner of" relationship; we're using http://chemheritage.org/ns/mannerOf instead.
Fixes #1122


jrochkind:
Nice work @eddierubeiz , so many places to add the new field, eh?

@HKativa note that we _are_ adding this to the "creators" facet, as "after" was already in there.

PR #1126, opened 2018-07-03
If the user enters an impossible date range, swap the dates.

### LINKED ISSUE #1125: Date range dropdown: check that start date is before end date before running search.
Not sure what's up with this one, but it's got a referrer of an item page, so it may be a URL generated by our app itself causing an error, which is more troublesome than some human or bot constructed external URL causing an error. 

## Backtrace

    line 222 of [PROJECT_ROOT]/app/controllers/catalog_controller.rb: index

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/53196/faults/38442607)

eddierubeiz:
Fixes #1125 . We're already subclassing catalog#index , so I'm just swapping the parameters before we call super.

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -219,7 +219,20 @@ class CatalogController < ApplicationController
   end
 
   def index
+    @_params = swap_dates_if_needed(@_params)
     super
     @parent_presenter_lookup = parent_lookup_hash(@document_list)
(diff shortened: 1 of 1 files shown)


jrochkind:
consider whether a fix for this could/should be submitted as a PR for blacklight_range_limit. Might look different than the fix here.

PR #1129, opened 2018-07-09
use new-to-rails5 config.assets.quiet=true for somewhat less overwhelming dev logs

jrochkind:


jrochkind:
@eddierubeiz this is a cool tool that lets you see differences in the generated Rails apps (with just `rails new`) for different Rails versions. 

In this one, I verified that the line being added here is actually generate by default in Rails 5 and onward. 

http://railsdiff.org/4.2.10/5.0.0

PR #1130, opened 2018-07-10
wip PDF originals

### LINKED ISSUE #994: Managing works that are PDFs
Our first PDF use case is to include the [Neville rare book catalog](http://othmerlib.chemheritage.org/search~S6?/Xneville+catalog&searchscope=6&SORT=D/Xneville+catalog&searchscope=6&SORT=D&SUBKEY=neville+catalog/1%2C7%2C7%2CB/frameset&FF=Xneville+catalog&searchscope=6&SORT=D&1%2C1%2C) in the Digital Collections. Jim is hoping that the text will be indexed by Google. He would like people to be able to download the PDF and perhaps view it in their browser. We understand that it won't work with our current image viewer.

- [x] Thumbnail logic needs review/feedback possibly (or not) including from stakeholders and/or users
    * decided current first page of PDF is good enough for now, Neville catalog will have new title pages created. 

## Maybe:

- ( ) re-enable zip download for multi-item works including pdfs *(not for now, we might not have lots of multi-PDF things)*

- [x] add page count in download links for PDFs? YES

- [x] Should the fact that the original is a PDF (and it's page count) be displayed somewhere else on page? YES, add file type to metadata

- [x] Make sure PDF originals are in index XML files given to google/google images? (yes)

- ( ) Neville Catalog PDF is really large file size; spend time figuring out why and if it can be reduced? *(will handle later, doesn't need to hold up this pr)*

jrochkind:
closes #994

### DIFF
--- a/app/assets/javascripts/chf_image_viewer.js
+++ b/app/assets/javascripts/chf_image_viewer.js
@@ -566,9 +566,20 @@ jQuery(document).ready(function($) {
 
     $(document).on("click", "*[data-trigger='chf_image_viewer']", function(event) {
       event.preventDefault();
+
       var id = this.getAttribute('data-member-id');
       chf_image_viewer().show(id);
     });
+
+    // Links set up for things that are not images, we just display them in the browser, using
+    // href supplied.
+    $(document).on("click", "*[data-trigger='chf_view_original']", function(event) {
+      event.preventDefault();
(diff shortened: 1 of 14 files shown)


jrochkind:
Needs reindex of works and filesets for page counts.

PR #1131, opened 2018-07-17
update solr_wrapper to 2.0

jrochkind:
To be able to succesfully download solr from apache or apache mirrors. 
https://groups.google.com/forum/#!topic/samvera-tech/sOnSa2NEZQg


jrochkind:
Will also allow us to use new solr_wrapper `mirror_url` configuration, possibly useful to deal with apache rate-limiting travis. But at the moment the princeton mirror was itself misbehaving, so I didn't configure it.

PR #1132, opened 2018-07-19
WIP Convert between filesets and child works

eddierubeiz:
Allows the user to convert between child works and files.

**Constraints**
- Only one item can be converted at a time. I've put some work into doing several items at a time, but that functionality isn't on this branch; we want to get this working before embarking on that more ambitious project.
- The order of items in the parent work is preserved.
- Thumbnail and "representative" item is preserved.
- Converting a child work to a file deletes metadata, but the user is warned before the operation starts.
- To prevent corruption, checks are run on the relationship between the item and its parent before the conversion is run.
- The user interface is pretty simple - we're just adding buttons on the "File Detail" and item record pages.

**To do**
- This feature is not yet in the tests.
- Check that the user has rights to edit all items concerned before running the conversion.
- If an item is a member of multiple parent works, the behavior is as-yet undefined.


jrochkind:
Another To Do: Copy some metadata from parent when converting from file set to child work?

PR #1137, opened 2018-07-27
WIP Allow the user to convert between child works and files

eddierubeiz:
Allows the user to convert between child works and files.

**Features**

- Only one item can be converted at a time. I've put some work into doing several items at a time, but that functionality isn't on this branch; we want to get this working before embarking on that more ambitious project.
- The order of items in the parent work is preserved.
- Thumbnail and "representative" status are preserved.
- Converting a child work to a file deletes metadata, but the user is warned before the operation starts.
- We run a bunch of checks, both client and server side, before the conversion begins.
- The user interface is pretty simple - we're just adding buttons on the "File Detail" and item record pages.
- Converting from a fileset to a child work takes time, but is not substantially slower than creating a GenericWork would be in other circumstances, even when the fileset is part of a work that has several hundred other filesets and dozens of GenericWorks in its member set.


eddierubeiz:
This is substantially complete but I would like to wait until Hillary and any other interested parties can take a look at it in Staging, especially using files with complex metadata attached to large works. (The obvious test case would be Ramelli.)

eddierubeiz:
Converting from a fileset to a child work takes time, but is not substantially slower than creating a GenericWork would be in other circumstances, even when the fileset is part of a work that has several hundred other filesets and dozens of GenericWorks in its member set.

app/controllers/member_conversion_controller.rb
```diff
@@ -0,0 +1,202 @@
...
+    if !validate_for_switch_to_child_work(parent_work, file_set)
+      flash[:notice] = "\"#{file_set.title.first}\" can't be promoted to a child work."
+      redirect_to "/concern/parent/#{parent_work.id}/file_sets/#{file_set.id}"
+      return
+    end
+    place_in_order = parent_work.ordered_members.to_a.find_index(file_set)
+    transfer_thumbnail = (parent_work.thumbnail == file_set)
+    transfer_representative = (parent_work.representative == file_set)
+    new_child_work = GenericWork.new(title: file_set.title)
+    acquire_lock_for(parent_work.id) do
```
jrochkind: I wonder if we can wrap the `acquire_lock_for` around _less_ stuff?  

As this stuff is so slow to happen... for the entire extent of the acquire_lock_for, nobody else will be able to edit the parent, and in fact their browsers (and the rails worker thread) will just hang waiting for this one to finish. 

acquire_lock_for only really needs to be around any mutations to the work. So I think everything ot line 29 (just creating the child) doesn't need to be in it. Neither does `set_thumbnail_and_rep(new_child_work, file_set, true, true)`, or `new_child_work.save!`.  Just things that might touch the parent. 

Which granted can be confusing to know what things will end up tryign to save the parent (leaky abstractions). 

I might not be totally right, please think it through and/or investigate if needed rather than just follow my directions... but I think it might pay off to try to minimize what's inside the `acquire_lock_for`, as longly-held locks are such a mess for everything.

app/controllers/member_conversion_controller.rb
```diff
@@ -0,0 +1,202 @@
+class MemberConversionController < ApplicationController
+  include CurationConcerns::Lockable
+
+  #Promote a FileSet to a child GenericWork.
+  def to_child_work
+    begin
+      parent_work = GenericWork.find(params['parentid'])
+      file_set = FileSet.find(params['filesetid'])
+    rescue
```
jrochkind: No bare `rescue` please, rescue the proper class. I think in this case some kind of `NotFound` exception, not sure which one?

eddierubeiz: Oops - thanks for catching that.

app/controllers/member_conversion_controller.rb
```diff
@@ -0,0 +1,202 @@
...
+        new_child_work.save!
+        copy_metadata_from_parent(parent_work, new_child_work)
+        parent_work.representative_id = nil if transfer_representative
+        parent_work.thumbnail_id = nil if transfer_thumbnail
+        remove_member_from_parent(parent_work, file_set)
+        parent_work.save!
+        add_member_to_parent(new_child_work, file_set, 0)
+        set_thumbnail_and_rep(new_child_work, file_set, true, true)
+        set_thumbnail_and_rep(parent_work, new_child_work, transfer_thumbnail, transfer_representative)
+        new_child_work.save!
```
jrochkind: Also not sure if there's a way to re-organize these so you don't need to save the parent work and child work _twice_ each -- we know saves are expensive. Can it be reordered so each only needs to be saved once?

There's a lot of things going on here -- I think they all really need to go  on -- but maybe re-organizing in into "paragraphs" with a little comment header explaining what's going on would help. Not totally sure. I don't totally follow the `transfer_representative` stuff though, maybe a brief comment explaining why you only sometimes need to do this.

eddierubeiz:
Promotion and demotion both take roughly 5 minutes for members of a large work. See e.g. https://staging.digital.sciencehistory.org/concern/parent/4b29b614k/file_sets/g732d9153

jrochkind:
5 minutes!?! Any sense of where that time is going, within the 5 minutes?  Just saving the parent work with many members, maybe?

eddierubeiz:
I'll look into it.
Also, if you want to try refactoring this, keep in mind that spec/controllers/member_conversion_spec.rb yesterday gave me false "pass" results; e.g. it would pass, and then later fail, given what appeared to me the same inputs -- to wit, the same files, ./bin/spring stop run before each test run.
However, a "fail" result always indicated that something was wrong.

jrochkind:
Info from Tom Johnson on yet another thing we could try:

-----------------

It's true that Fedora itself takes a major performance hit for accessing resources with many outgoing internal references, but I believe the issue Johnathan has run into with accessing/updating associations in ActiveFedora is several orders of magnitude worse than this. Specifically, the cost of this access scales linearly twice: first taking that performance hit for accessing the initial resource, then it sends additional GET requests for each of the items in the association, incurring the Fedora-side costs *again*, as well as network latency, and any ruby-side costs of allocating these objects. For the simple add-a-resource-to-the association case, data from this second group of requests will never be used. In many other cases, these costs can be deferred and amortized by retrieving data just-in-time.

Johnathan: If your goal is to end-around ActiveFedora to add an URI reference to a property, you might actually be able to do this without too much pain. The rough approach I have in mind is:

1. Add the URI to the property's underlying ActiveTriples representation: `my_work.resource.ordered_members << other_work`
2. Manually mark the property as changed; something like: `my_work.attribute_will_change!(:ordered_members)` (???)
3. Save as usual: `my_work.save`

If you run into problems with this approach, ActiveFedora::ChangeSet, may be the best place to begin troubleshooting. If you can get this module returning the set of ordered members URIs you are looking for without loading all the members, you should be able to use the usual save method. https://github.com/samvera/active_fedora/blob/master/lib/active_fedora/change_set.rb

- Tom

jrochkind:
initial attempts to try something along those lines got me _nowhere_, even his step 1 doesn't work anything like that in actuality. :shrug:

eddierubeiz:
Fixes https://github.com/sciencehistory/chf-sufia/issues/1024

PR #1143, opened 2018-08-21
WIP Dont show the GDPR cookie banner to the tests

### LINKED ISSUE #1139: Cookie accept banner breaks tests
The GDPR cookie banner is causing front-end tests to break in unpredictable and annoying ways. 
(see https://github.com/sciencehistory/chf-sufia/pull/1067)
I would like to disable it for *most* front-end tests.

eddierubeiz:
Set Rails.application.config.hide_accept_cookies_banner in the test config file, then check it in the template.
I also removed three references to the banner from the tests.
Fixes #1139 .


app/views/shared/_accept_cookies_banner.html.erb
```diff
@@ -1,26 +1,33 @@
...
-            </div>
-            <div class="col-sm-2 i-accept-div text-center">
-                <a href="#"
-                    class="i-accept-link"
-                    aria-label="dismiss cookie message"
-                    role="button"
-                    tabindex="0">I Accept</a>
+<%
+    if Rails.application.config.respond_to?(:hide_accept_cookies_banner) &&
+    Rails.application.config.hide_accept_cookies_banner
```
jrochkind: Is there any reason it wouldn't, any reason to check `respond_to?`?  It's a function of our app that it always has this once we've added it to config/application.rb, I think?

app/views/shared/_accept_cookies_banner.html.erb
```diff
@@ -1,26 +1,33 @@
...
+%>
+    <%# Don't show this cookie banner to the unit tests unless they specifically request it.%>
+<% else %>
+    <nav class="navbar navbar-default navbar-fixed-bottom accept_cookies_banner_nav">
```
jrochkind: no need for an `if` with an empty body followed by `else` -- just use `unless` or `if !`.

app/views/shared/_accept_cookies_banner.html.erb
jrochkind: oh I see, you've ONLY put `config.` in the test environment.  I'd add it to `config/application.rb` as false, then override in environments/test.rb as true.

jrochkind:
We're not too consistent about where we store global config state like this. We also use the CHF::Env class, which allows setting via ENV or a config file... but I guess doesn't easily allow resetting during program execution, it assumes state never changes once program is launched. Sometimes worked around in tests with mocking. 

I have no idea where the 'right' place to put this state is really, just thought I'd mention the alternative, I think it's fine to leave it here too, we already weren't always consistent. 

Excited about having this in there. :) Not sure why one of the travis builds is failing, maybe just a capybara race condition?

jrochkind:
ready for merge? looks good to me, feel free to click merge. (I would have but for 'wip' still in title).

config/application.rb
```diff
@@ -220,6 +220,11 @@ class Application < Rails::Application
     # Active job should use resque
     config.active_job.queue_adapter = :resque
 
+    # Show the GDPR "I accept" cookies banner by default.
+    # This setting is overridden in test.rb,
+    # so the tests don't have to click "I accept".)
+    config.hide_accept_cookies_banner = true
```
eddierubeiz: This is a bug -- it should actually be false. We do not want to hide the cookie banner by default, we only want to hide it for the unit tests. Fixed in https://github.com/sciencehistory/chf-sufia/pull/1155 .

PR #1146, opened 2018-08-29
Extra blacklight validation

eddierubeiz:
These tweaks avoid two recent errors we've gotten in production by checking search inputs before they are sent to blacklight_range_limit. Fixes https://github.com/sciencehistory/chf-sufia/issues/1144
Note: I'm also going to suggest a fix to the actual blacklight_range_limit project based on https://github.com/sciencehistory/chf-sufia/pull/1126 .

### DIFF
--- a/app/controllers/collections_show_controller.rb
+++ b/app/controllers/collections_show_controller.rb
@@ -23,6 +23,10 @@ class CollectionsShowController < CatalogController
   # so :id can continue being our parent collection id.
   # displays values and pagination links for a single facet field
   def facet
+    unless params.key?(:facet_id)
+      redirect_back fallback_location: { action: "index", id: params['id'] }
+      return
+    end
     @facet = blacklight_config.facet_fields[params[:facet_id]]
     @response = get_facet_field_response(@facet.key, params)
     @display_facet = @response.aggregations[@facet.key]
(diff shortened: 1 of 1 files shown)


eddierubeiz:
Thinking about this a bit more, I don't think it's a good idea to submit a PR to swap out-of-order range limits within the blacklight_range_limit code. I think it's fine for us to decide to do that within chf-sufia, of course, but within blacklight_range_limit it feels surreptitious to me.

I also just discovered the code ships with a [test](https://github.com/projectblacklight/blacklight_range_limit/blob/46987fcac31942ef5d24da5d06b83056107840b1/spec/lib/blacklight_range_limit/segment_calculation_spec.rb#L39 
) that codifies the existing ([fixed](https://github.com/projectblacklight/blacklight_range_limit/commit/8e6cb2f280b6d8022eb3112954f3724e0ef3abac#diff-f30070d246658df27708487cbfda99a2), from their point of view) functionality.

jrochkind:
Gah, travis failure:

```
     Capybara::Poltergeist::MouseEventFailed:
       Firing a click at co-ordinates [893, 520.5] failed. Poltergeist detected another element with CSS selector 'html.fixedsticky-withoutfixedfixed.wait body.localized-layout.branded-body-font.modal-open div.modal-backdrop.fade' at this position. It may be overlapping the element you are trying to interact with. If you don't care about overlapping elements, try using node.trigger('click').
```

I think this is the thing we thought was fixed involving the cookie banner maybe. It might be something else, but can you confirm this branch includes that fix, and/or rebase/merge master to ensure it does?

I did click "rebuild" in travis, maybe it'll pass trying again.

eddierubeiz:
Will do!

PR #1150, opened 2018-09-05
Fix for bug with batch edit of copyright status.

eddierubeiz:
Fixes https://github.com/sciencehistory/chf-sufia/issues/1059.

Our subclassed model_attributes(params) in batch_edit_form.rb was converting the value of params[:rights] from a string to an array, but it did so *after* calling super instead of *before*.

The test I wrote back in the spring now passes, so I'm removing the :skip .


jrochkind:
nicely done! ready for review/merge?

eddierubeiz:
Yeah - go ahead! :)

eddierubeiz:
Will investigate test failure later today. (Passes in dev.)

PR #1153, opened 2018-09-07
(DO NOT MERGE) Solr wrapper test fix

eddierubeiz:
Monkey patch to incorporate https://github.com/cbeer/solr_wrapper/pull/119/commits/eaab9cd5bba2b5d3bc94f33c15cee089462089ae into our code until/unless solr_wrapper incorporates it.
(We can also simply not use the Princeton mirror for a bit, or just set validate: false in the SOLR yaml ).

### DIFF
--- a/config/solr_wrapper_test.yml
+++ b/config/solr_wrapper_test.yml
@@ -11,6 +11,11 @@ collection:
     dir: solr/config
     name: hydra-test
 version: 6.3.0
+
+# To simply ignore the checksum, you can set validate to false.
+# The default is true; we are currently downloading the checksum.
+# validate: false
+
 # Attempt to use mirror url to keep travis tests from being blocked by
 # apache for too many solr downloads.
 mirror_url: http://lib-solr-mirror.princeton.edu/dist/
--- a/lib/solr_wrapper_patch.rb
+++ b/lib/solr_wrapper_patch.rb
@@ -1,0 +1,23 @@
+# Hack to monkey-patch solr_wrapper to get the checksum
+# file from archive.apache.org instead of www.us.apache.org .
+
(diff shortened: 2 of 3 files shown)


config/solr_wrapper_test.yml
```diff
@@ -11,6 +11,7 @@ collection:
     dir: solr/config
     name: hydra-test
 version: 6.3.0
+# validate: false # until solr_wrapper stops hardwiring the checksum URL
```
jrochkind: this is commented out, you still want to put it in there commented out?

jrochkind:
Another alternative is simply removing our `mirror_url` config. It wasn't generally required, only occasionally when apache.org rate-limits travis -- and we discover the `mirror_url` feature is actually broken in solr_wrapper. Although we can apparently fix it with this monkey patch. 

If we remove the mirror_url, we could spend more time trying to get cbeer to merge/release the fix in solr_wrapper, or convincing others that we should fork it so someone else can.

config/solr_wrapper_test.yml
eddierubeiz: It should be commented out, yes, but the comment is misleading. Let me amend the commit using my new git --amend superpowers.

eddierubeiz: Updated and off to lunch. BRB

jrochkind:
I guess we'll do this for now, although honestly it kind of terrifies me.

jrochkind:
The best alternative would be for the already-existing patches in solr_wrapper to be merged/released. I filed a ticket on solr_wrapper asking if cbeer would consider adding additional committers. https://github.com/cbeer/solr_wrapper/issues/122

PR #1164, opened 2018-10-02
Get solr from s3

### LINKED ISSUE #1163: Download SOLR from a local URL
Amend solr_wrapper so it can download SOLR from a local URL. The URL should be accessible to us (both locally and on Travis) but not to anyone else.

eddierubeiz:
Fixes #1163 .
Daniel created an S3 account (travis) which can only LIST and GET objects from the science-history-build-cache bucket.
We declare two env variables in .travis.yml that define AWS credentials for that account, and set the solr zip download url explicitly in solr_wrapper_test.yml .
We also set validate to false.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,6 @@ services:
   - postgresql
 
 env:
-  matrix:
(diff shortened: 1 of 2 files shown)


eddierubeiz:
@jrochkind recommends "add a comment (`#` comments in .travis.yml should be fine) with link to the travis instructions for encrypted env, and a list of what the actual variables are in plaintext. (I don’t think we need to hide the variable names, just the values"

PR #1167, opened 2018-10-03
Get solr from Princeton mirror; do not check the checksum.

### LINKED ISSUE #1166: Download SOLR from Princeton, but don't check the checksum file

eddierubeiz:
Use the Princeton mirror url to keep travis tests from being blocked by apache for too many solr downloads.

Ignore the checksum for now.

When the checksum can _also_ be fetched from Princeton, which will happen when https://github.com/cbeer/solr_wrapper/pull/119 is merged, we'll turn validate back on.

See also:
https://github.com/cbeer/solr_wrapper/issues/122
https://github.com/sciencehistory/chf-sufia/pull/1153

Fixes #1166.

### DIFF
--- a/config/solr_wrapper_test.yml
+++ b/config/solr_wrapper_test.yml
@@ -13,9 +13,14 @@ collection:
 version: 6.3.0
 
 
-# Attempt to use mirror url to keep travis tests from being blocked by
+# Use mirror url to keep travis tests from being blocked by
 # apache for too many solr downloads.
-# (Commenting this out, as it was more trouble than it was worth. See 
+# Ignore the checksum, which is coming from apache until
+# https://github.com/cbeer/solr_wrapper/pull/119 is merged.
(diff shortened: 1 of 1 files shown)


jrochkind:
How the heck are we now passing CI without having merged this?  I forget the state of this PR, should we still merge it?

eddierubeiz:
We don't need this after all! We can come back to it as needed.

jrochkind:
We can't quite remember why we thought we needed this -- we were having troubles with the travis build being able to download solr, involving apache rate-limiting, trying to use mirror url to get around that, and then apparent bugs or lack of feature in the solr_wrapper `mirror_url` feature. 

We aren't currently having any trouble with the build though. Maybe just because apache isn't currently rate limiting us? 

We can revisit this if the problem happens again and we are refreshed as to context.

PR #1182, opened 2018-11-08
Using https for OPAC (Sierra) links

eddierubeiz:
Fixes https://github.com/sciencehistory/chf-sufia/issues/1178 .

### DIFF
--- a/app/presenters/chf/collection_show_presenter.rb
+++ b/app/presenters/chf/collection_show_presenter.rb
@@ -19,12 +19,14 @@ module CHF
 
     def urls_to_catalog
       return [] unless @solr_document.related_url.present?
-      @urls_to_catalog ||= @solr_document.related_url.find_all do |url|
-        url.start_with?("http://othmerlib.chemheritage.org/record=") || url.start_with?("http://othmerlib.sciencehistory.org/record=")
-      end
+      @urls_to_catalog ||= @solr_document.related_url.
+        find_all { |url| is_opac_url?(url) }
     end
 
-
+    def is_opac_url?(url)
+      test_regex = /^https?:\/\/othmerlib\.(chemheritage|sciencehistory)\.org\/record=/
+      test_regex.match(url).nil? ? false : true
+    end
 
   end
 end
--- a/app/views/catalog/_search_sidebar.html.erb
+++ b/app/views/catalog/_search_sidebar.html.erb
@@ -14,7 +14,7 @@ https://github.com/projectblacklight/blacklight/blob/v6.7.2/app/views/catalog/_s
       <li class="list-group-item"><%= link_to "https://www.sciencehistory.org/search/site/#{Addressable::URI.encode_component(params[:q], Addressable::URI::CharacterClasses::PATH)}" do %>
         <span class="fa fa-search"></span>  Science History Institute
       <% end %></li>
-      <li class="list-group-item"><%= link_to "http://othmerlib.sciencehistory.org/search/?searchtype=X&SORT=D&searcharg=#{Addressable::URI.encode_component(params[:q], Addressable::URI::CharacterClasses::QUERY)}" do %>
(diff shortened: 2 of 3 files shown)


eddierubeiz:
@yonyitz and @HKativa: this is now in staging. Compare the "View in Library catalog" links in 
https://staging.digital.sciencehistory.org/works/4b29b614k (https) and 
https://digital.sciencehistory.org/works/4b29b614k .

app/presenters/chf/collection_show_presenter.rb
```diff
@@ -20,7 +20,8 @@ def needs_permission_badge?
     def urls_to_catalog
       return [] unless @solr_document.related_url.present?
       @urls_to_catalog ||= @solr_document.related_url.find_all do |url|
-        url.start_with?("http://othmerlib.chemheritage.org/record=") || url.start_with?("http://othmerlib.sciencehistory.org/record=")
+        url.include?("othmerlib.chemheritage.org/record=") ||
+        url.include?("othmerlib.sciencehistory.org/record=")
       end
```
jrochkind: While this is _probably_ good enough, technically a URL _could_ include this string without it actually being the hostname. Probably unlikely, but say `http://example.org/othmerlib.chemheritage.org/record=/somethingelse` is technically a valid url. 

That's why i was doing the `starts_with`,  but I see that had a bug, repeating the same condition twice! It was meant to be `url.start_with?("http://othmerlib.chemheritage.org/record=") || url.start_with?("https://othmerlib.sciencehistory.org/record=")`  (the second one being https). 

Could also be expressed as a regexp `%r{\Ahttps?//othermerlib\.chemheritage\.org/record=}`.  Don't know if tha'ts more or less legible/prone to error. 

Do you think it's worth correcting to make sure it's only matching URLs that really are "start with" rather than the unlikely but possible ones like the above example?

app/views/catalog/_search_sidebar.html.erb
```diff
@@ -14,7 +14,7 @@ https://github.com/projectblacklight/blacklight/blob/v6.7.2/app/views/catalog/_s
       <li class="list-group-item"><%= link_to "https://www.sciencehistory.org/search/site/#{Addressable::URI.encode_component(params[:q], Addressable::URI::CharacterClasses::PATH)}" do %>
         <span class="fa fa-search"></span>  Science History Institute
       <% end %></li>
-      <li class="list-group-item"><%= link_to "http://othmerlib.sciencehistory.org/search/?searchtype=X&SORT=D&searcharg=#{Addressable::URI.encode_component(params[:q], Addressable::URI::CharacterClasses::QUERY)}" do %>
+      <li class="list-group-item"><%= link_to "https://othmerlib.sciencehistory.org/search/?searchtype=X&SORT=D&searcharg=#{Addressable::URI.encode_component(params[:q], Addressable::URI::CharacterClasses::QUERY)}" do %>
```
jrochkind: i had totally forgotten about this one, thanks for being thorough and looking for em all!

app/presenters/chf/collection_show_presenter.rb
eddierubeiz: Absolutely. Will include a comment.

app/views/catalog/_search_sidebar.html.erb
eddierubeiz: The magic of grep :)

app/presenters/chf/collection_show_presenter.rb
jrochkind: oh wait, it was one sciencehistory and one chemheritage originally! I misread my own code, ha. Yeah, probably wroth changing to a regexp, or two regexps, if we want to make it more precise.

jrochkind: Alternately, could be done with a `URI.parse` and checking hostname and path components independently (have to make sure to rescue parse errors). 

Or maybe it's good enough how it is, but it bothers me that it technically, if unlikely, could match a URL that it shouldn't. :)

eddierubeiz: I think that's perfectly reasonable.

PR #1185, opened 2018-11-08
Use vipsthumbnail to create PDF thumbnails

eddierubeiz:
This is ready for a next round of review.
Vipsthumbnail + vips sharpen create derivatives comparable to the ImageMagick ones.
Need to verify that these do in fact reduce load significantly on the server when used in batch PDF ingests. (They should.)
I can imagine spending more time to remove some repetition in the thumbnail creation code by creating a couple functions, but not necessarily as part of this PR.

### DIFF
--- a/app/assets/stylesheets/local/work-show.scss
+++ b/app/assets/stylesheets/local/work-show.scss
@@ -184,10 +184,9 @@
       max-height: 300vw;
 
       &.pdf_border {
-        border: 2px solid grey;
+        border: 2px solid #050939;
       }
     }
-
 
     .show-page-member-title {
       position: absolute;
--- a/app/assets/stylesheets/local/work-show.scss
+++ b/app/assets/stylesheets/local/work-show.scss
@@ -182,7 +182,12 @@
       // aspect ratio, cause the dimensions are so responsive. This is just
       // a rough huge limit, if it's not working out just delete this and give up.
       max-height: 300vw;
+
+      &.pdf_border {
+        border: 2px solid grey;
+      }
     }
+
 
     .show-page-member-title {
       position: absolute;
--- a/app/assets/stylesheets/local/work-show.scss
+++ b/app/assets/stylesheets/local/work-show.scss
@@ -182,10 +182,6 @@
       // aspect ratio, cause the dimensions are so responsive. This is just
       // a rough huge limit, if it's not working out just delete this and give up.
       max-height: 300vw;
-
-      &.pdf_border {
-        border: 2px solid #050939;
-      }
     }
 
     .show-page-member-title {
--- a/app/helpers/image_service_helper.rb
+++ b/app/helpers/image_service_helper.rb
@@ -88,12 +88,6 @@ module ImageServiceHelper
       }
     }
 
-    if member.respond_to? 'representative_content_type' and
-      !member.representative_content_type.blank? and
-      member.representative_content_type.include? "pdf"
-      args[:class] += ['pdf_border']
-    end
-
     src_args = member_src_attributes(member: member, size_key: size_key)
 
     if lazy
--- a/app/helpers/image_service_helper.rb
+++ b/app/helpers/image_service_helper.rb
@@ -88,7 +88,11 @@ module ImageServiceHelper
       }
     }
 
-    args[:class] += ['pdf_border'] if member.representative_content_type.include? "pdf"
+    if member.respond_to? 'representative_content_type' and
+      !member.representative_content_type.blank? and
+      member.representative_content_type.include? "pdf"
+      args[:class] += ['pdf_border']
+    end
 
     src_args = member_src_attributes(member: member, size_key: size_key)
 
--- a/app/helpers/image_service_helper.rb
+++ b/app/helpers/image_service_helper.rb
@@ -88,6 +88,8 @@ module ImageServiceHelper
       }
     }
 
+    args[:class] += ['pdf_border'] if member.representative_content_type.include? "pdf"
+
     src_args = member_src_attributes(member: member, size_key: size_key)
 
     if lazy
--- a/app/services/chf/create_derivatives_on_s3_service.rb
+++ b/app/services/chf/create_derivatives_on_s3_service.rb
@@ -151,7 +151,6 @@ module CHF
     # it to helper methods, definitely not thread-safe, don't share instances
     # between threads, you weren't going to anyway.
     def call
-
       futures = []
 
       # mktmpdir will clean up tmp dir and all it's contents for us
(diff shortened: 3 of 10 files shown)


eddierubeiz:
Here is a comparison of 3 recipes for making a thumbnail out of the same PDF on staging:

Imagemagick:

	User time (seconds): 1.70
	System time (seconds): 0.13
	Elapsed (wall clock) time (h:mm:ss or m:ss): 0:01.67
	Maximum resident set size (kbytes): 440964
	Average resident set size (kbytes): 0
	
Current Vips recipe
	User time (seconds): 0.04
	System time (seconds): 0.00
	Elapsed (wall clock) time (h:mm:ss or m:ss): 0:00.04
	Maximum resident set size (kbytes): 13484
	Average resident set size (kbytes): 0

Vips again but with jpg quality set to 98 instead of 85
	User time (seconds): 0.05
	System time (seconds): 0.00
	Elapsed (wall clock) time (h:mm:ss or m:ss): 0:00.05
	Maximum resident set size (kbytes): 14008
	Average resident set size (kbytes): 0

jrochkind:
Vips get san order of magnitude RAM improvement, which was our goal, so that's great. 

Have you figured out the perceivable image quality issue yet? I don't think the IM recipe was using Q95... but somehow it was making a lot better quality thumbs. I remember fighting with it with IM, but I can't remember the details, I can look at the code again and try to remember what was going on, if you're stuck.

sanfordd:
This may be too obvious to be helpful, but have we checked the DPI settings as well?

jrochkind:
Aha, I left myself some hints in comments. While I didn't entirely understand what was going on, the key arguments to IM seemed to be `-density 400`, and doing  `-unsharp 0x3.0`. 

I don't know if there are equivalents in vips. 

Dan, I don't believe there is anythign as simple as "a DPI setting" that is relevant here. There is DPI metadata in a JPG, but I believe it's generally ignored by browsers, and I think the issue here is the actual JPG quality, not DPI metadata. The pixels are the pixels. (Yes, I've spent a lot of time in the last year trying to learn more about digital imaging :) )

eddierubeiz:
Vips does sharpening and blurring, yeah. It would be more complicated, though, so let me just see what improving the quality to 98 looks like in staging.

eddierubeiz:
See https://staging.digital.sciencehistory.org/catalog?search_field=all_fields&q=Babson for a comparison. I can definitely look into using VIPS for sharpening.

eddierubeiz:
https://stackoverflow.com/questions/35603961/how-to-translate-imagemagicks-unsharpmask-settings-to-vipss-vips-sharpen-set

app/assets/stylesheets/local/work-show.scss
```diff
@@ -182,6 +182,10 @@
       // aspect ratio, cause the dimensions are so responsive. This is just
       // a rough huge limit, if it's not working out just delete this and give up.
       max-height: 300vw;
+
+      &.pdf_border {
+        border: 2px solid #050939;
+      }
```
jrochkind: should we define a custom CSS mixin, so we don't need to repeat the border definition in two places?  Or find another way to not need to repeat the border definition in two places? Low priority, not a big deal if we keep repeating.

app/helpers/image_service_helper.rb
```diff
@@ -88,6 +88,12 @@ def member_image_tag(parent_id:, member:, size_key: nil, lazy: false)
       }
     }
 
+    if member.respond_to? 'representative_content_type' and
+      !member.representative_content_type.blank? and
+      member.representative_content_type.include? "pdf"
+      args[:class] += ['pdf_border']
```
jrochkind: I think this should probably be `== 'application/pdf'` instead of `include? 'pdf'`; I think the latter is vague for no reason.

app/views/curation_concerns/generic_works/_generic_work.html.erb
```diff
@@ -14,9 +14,7 @@
 
 <li id="document_<%= generic_work.to_param %>" class="chf-results-list-item document <%= render_document_class generic_work %>" itemscope itemtype="<%= generic_work.itemtype %>">
     <% doc_presenter = Chf::IndexPresenter.new(generic_work, self) %>
-
-
-    <div class="chf-results-list-item-thumb">
+    <div class="chf-results-list-item-thumb <%= "pdf_border" if doc_presenter.is_pdf? %>">
       <%# render_thumbnail_tag generic_work %>
```
jrochkind: deciding whether to include `pdf_border` class in two completely different ways in two different parts of code. That may just be necessary (or really inconvenient to do different) because of how sufia is architected though? I haven't completely wrapped my head around it, but worth at least considering if this can be DRY'd in some non-awful way.

app/services/chf/create_derivatives_on_s3_service.rb
```diff
@@ -301,45 +301,78 @@ def create_image_jpg_derivative_future(width:, type_key:)
       end
     end
 
-
-    # Takes a pdf, makes a thumb. Includes small border around thumb, since this
-    # works better for typical white-bg print-like PDFs.
-    def create_pdf_jpg_thumb_future(width:, type_key:)
+    def create_pdf_thumb_future(width:, type_key:)
       defn = get_type_defn(type_key)
```
jrochkind: any reason you changed the name from `create_pdf_jpg_thumb_future`?  I had `jpg` in there to indicate that the output is a "jpg", and because other similar methods in this case include "jpg" in their names for same.

eddierubeiz: Nah, this is just an artifact. Will change back.

app/services/chf/create_derivatives_on_s3_service.rb
```diff
@@ -301,45 +301,78 @@ def create_image_jpg_derivative_future(width:, type_key:)
...
       style = defn.style
+
       output_path = Pathname.new(working_dir).join(type_key.to_s).sub_ext(defn.suffix).to_s
+      tmp_output_path = output_path.sub('.jpg', "_tmp.jpg")
+
```
jrochkind: This adds a `tmp_output_path` because of the fact that you're now running multiple vips commands?

We need to make sure any temporary files are cleaned up and don't remain on disk after thumb is created. Otherwise bulk creation of thumbs can fill up disk space. This actually applies to the "final" thumb (on local disk) too. I haven't totally wrapped my head around what's going on or how this was handled in previous code. Using the ruby stdlib Tempfile class may be helpful.

jrochkind: Actually as long as you are creating in `working_dir`, I think the surrounding code should take care of that, as the whole dir was created using ruby temp directory stuff. But maybe try to verify this is true?

jrochkind:
I think none of this had any automated tests before anyway, because it was such a pain to figure out how to do them?  If so, fine to continue with that.

app/services/chf/create_derivatives_on_s3_service.rb
eddierubeiz: Sounds reasonable. I feel like we were deleting the temp folder containing all our working files, but it's definitely worth triple-checking.

Also I'm not sure I really need tmp_output_path. I feel like vips can run these edits in place and save the edited image to the same file path. Will take a quick look.

PR #1196, opened 2018-11-30
Allow HTML in collection descriptions.

### LINKED ISSUE #1183: Add related URL label for Oral History Collection
Within Oral History Collection (https://digital.sciencehistory.org/collections/gt54kn818) please update "Related URL" to read "View Oral History catalog" instead.

eddierubeiz:
Sanitizes HTML on the way in and out, using the same mechanism used for GenericWork descriptions ( app/helpers/description_formatter_helper.rb ).
Fixes #1183 .


eddierubeiz:
Still need to add tests to this.

spec/features/collection_feature_spec.rb
```diff
@@ -11,11 +11,20 @@
...
     visit collection_path(collection)
 
     expect(page).to have_text("1 item")
     expect(page).to have_link(title, href: curation_concerns_generic_work_path(work.id))
 
+    # The factory (spec/factories/collections_factory.rb) sets the link to
+    # "<a href="https://en.wikipedia.org" target="_blank">Wikipedia</a>".
+    # The scrubber invoked in app/views/collections/_collection_description.erb
+    # is supposed to allow the link, but scrub the 'target' attribute.
+
```
jrochkind: The factory should just contain generic arbitrary data, not data we are testing against. 

Instead of editing the factory to include the value you want to test against with a comment to that effect, you should just explicitly specify the value you want to test against in the test. 

Up on line 11:

     let!(:collection) { FactoryGirl.create(:collection, :public, :with_image, decription: "whatever", members: [work]) }

PR #1197, opened 2018-12-03
WIP Use no-redirect content url for Tumblr.

eddierubeiz:
The Tumblr button, as currently implemented in production, apparently doesn't follow redirects in "content" urls, as currently in prod. My first attempt was to send a direct link to s3 instead. We gave this some thought and it didn't seem like a great idea in the long run.

I also discovered by accident that removing the "posttype: 'photo'" section of the share tool link fixes the problem: for some odd reason, when you omit that part of the link, Tumblr follows the redirect. Note that @jrochkind [specifically added the posttype in response to stakeholder request back in April](https://github.com/sciencehistory/chf-sufia/pull/1037), so we may or may not want to backtrack on an existing feature.

As of Tuesday this second fix is available for testing [in staging](https://staging.digital.sciencehistory.org/works/bc386j260). I'm going to wait until @HKativa gets back to do any more work on this, as I'm not that familiar with how Othermeralia uses our content (much less other users in the wild).

### DIFF
--- a/app/views/curation_concerns/base/_social_media.html.erb
+++ b/app/views/curation_concerns/base/_social_media.html.erb
@@ -23,16 +23,6 @@
     default_page_title
   end
 end %>
-
-<%
-  #HACK
-  #TODO: RENAME AND REFACTOR.
-  #This is just a proof of concept.
-  new_tumblr_url = member_src_attributes(member: presenter, size_key: :large )[:src]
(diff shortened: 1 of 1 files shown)


eddierubeiz:
Closing this; we're getting rid of the Tumblr button altogether after a discussion w/ stakeholders.

PR #1200, opened 2018-12-07
Remove tumblr button

### LINKED ISSUE #1171: Try to integrate PDFs with Tumblr
There's a research component to this; I imagine other comparable digital libraries have already solved this problem.

### LINKED ISSUE #1172: Share to Tumblr feature doesn't import image
Observed some wonky behavior with share to Tumblr feature starting October 30th:

From Chrome, I tried to share this work (single image) to Tumblr: https://digital.sciencehistory.org/works/bc386j260. When the screen opens up in Tumblr, it imports the metadata, but the image isn't showing up. Instead of the image, I get the generic menu for adding a photo via upload or the web.

Following the same steps in IE, the image appears when the screen opens up in Tumblr, but is replaced by the generic menu after logging into Tumblr to complete the post. 

Alternately tried to use the feature both when logged into Digital Collections site and when logged out; also when logged into Tumblr and when logged out with same result. Also tried deleting cookies in Chrome with same result. Also tested share feature with multi-image record with same result.

eddierubeiz:
I'm just commenting the button out using an "if false" block. In case Tumblr fixes their API, we can bring the button back in minutes.
Fixes #1171 
Fixes #1172


jrochkind:
I think we should probably just eliminate the code completely. While I sometimes have the temptation to leave things commented out too, best practices say leaving dead code commented out in your codebase just makes the codebase more confusing and is potential for bugs. In this case, guarded by an `if false` is probably even worse than just commented out. 

It'll still be there in the git history.

PR #1201, opened 2018-12-07
Adding oral history thumbnail

eddierubeiz:
Just adding the oral history thumbnail.


jrochkind:
Oh right, we have to then manually run a rake task to associate the image with a certain collection _after_ it's been deployed in the codebase. 

I initially forgot this, and wondered why it was just a lone binary not being used anywhere, til I remembered. (The collection thumb implementation isn't ideal, it was a stop-gap; we could improve it, if we weren't trying to rewrite the app instead).

PR #1203, opened 2018-12-10
update rails to avoid activejob CVE-2018-16476

jrochkind:


eddierubeiz:
This is marked "Pending", and has for a while. But both checks have actually passed:

https://travis-ci.org/sciencehistory/chf-sufia/builds/466106548
https://travis-ci.org/sciencehistory/chf-sufia/builds/466106734

PR #1205, opened 2018-12-12
Adding a target=_blank to description links

### LINKED ISSUE #1183: Add related URL label for Oral History Collection

eddierubeiz:
Uses nokogiri to add in the new link attribute.
Note that the target=_blank is not stored, but added at display time; the tests reflect this.
Fixes #1183

### DIFF
--- a/app/views/collections/_collection_description.erb
+++ b/app/views/collections/_collection_description.erb
@@ -1,5 +1,5 @@
 <% presenter.description.each do |description| %>
     <p class="collection_description">
-        <%= iconify_auto_link(format_description(description)) %>
+        <%=format_description(description)%>
     </p>
 <% end %>
--- a/app/helpers/description_formatter_helper.rb
+++ b/app/helpers/description_formatter_helper.rb
@@ -3,20 +3,25 @@ module DescriptionFormatterHelper
   def format_description(text, truncate: false)
     # sanitize. should have been sanitized on input, but just to
     # be safe.
-    text = DescriptionSanitizer.new.sanitize(text).html_safe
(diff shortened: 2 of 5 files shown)


app/views/collections/_collection_description.erb
```diff
@@ -1,5 +1,5 @@
 <% presenter.description.each do |description| %>
     <p class="collection_description">
-        <%= iconify_auto_link(format_description(description)) %>
```
eddierubeiz: No longer needed here -- we're treating both types of description the same way, which includes a pass through iconify_auto_link in the description formatter above.

jrochkind:
Looks good. 

It's a bit unfortunate that I think this parse with Nokogiri is probably a _second_ parse with Nokogiri, I think the `loofah` based sanitizer also does a nokogiri parse. But there is probably no great way to hook into it to do something different than loofah is doing. This is probably good enough.

jrochkind:
Hmm, hypothetically we could possibly write a custom loofah scrubber to hook into the single nokogiri parse. 

https://github.com/flavorjones/loofah#the-basics

I'm not sure what that code would look like, or how much of what's already there would have to be changed. Do you think it's worth spending more time on?

jrochkind:
@eddierubeiz I did it and pushed it as an additional commit. I'm doubting myself lately, not sure if this is an improvement, the opposite, or just shuffling deck chairs. Feedback welcome.

PR #1214, opened 2019-01-09
Fix oral histories citations

### LINKED ISSUE #1169: "Cite as" on Oral histories is not quite right
When we designed "Cite As", we had particular examplar/models, and we weren't doing OH yet. 

So eg https://digital.sciencehistory.org/works/gh93h041p

We show:

> Cite as: “Oral History Interview with William John Bailey,” June 3, 1986. https://digital.sciencehistory.org/works/gh93h041p."

The OPAC also has a manually entered suggested citation, and it differs/is better:

> Cite as: "William John Bailey, interviewed by James J. Bohing at the Univeristy of Maryland on 3 June 1986, (Philadelphia: Chemical Heritage Foundation, Oral History Transcript 0012.)"

It should not be hard to improve our citation display to be closer to that, if not exactly that. We have all of those data elements as machine-interprable data in the record, just need a special path to assemble them in the way we want for things marked Oral Histories (which they are as a 'genre' -- we already have genre-specific citation code). 

May also apply to RIS export improvements.

eddierubeiz:
closed #1169

### DIFF
--- a/app/models/chf/citable_attributes.rb
+++ b/app/models/chf/citable_attributes.rb
@@ -491,7 +491,7 @@ module CHF
       def title
         return work.title if work.interviewee.nil? || work.interviewer.nil?
         place = work.place_of_interview.nil? ? "" : "at #{work.place_of_interview.first}"
-        time = date.nil? ? "" : "on #{date.iso8601}"
+        time = original_date.nil? ? "" : "on #{original_date.strftime("%B %-d, %Y")}"
         "#{parse_name(work.interviewee.first)}, interviewed by #{parse_name(work.interviewer.first)} #{place} #{time}"
       end
 
@@ -516,7 +516,13 @@ module CHF
         'Science History Institute'.freeze
       end
 
+      # no date in CSL, we're embedding it in title instead
       def date
+        nil
+      end
+
+      # we don't want to be used in CSL, but we want to pull it out to use in title
+      def original_date
         return nil if work.date_of_work.length == 0
         begin
           date_recorded = Date.strptime(work.date_of_work.first, '%Y-%m-%d')
--- a/app/models/chf/citable_attributes.rb
+++ b/app/models/chf/citable_attributes.rb
@@ -36,10 +36,9 @@ module CHF
       @parent_work = parent_work.nil? ? work.parent_work_presenters.first : parent_work
       @edge_case_date_literals = !!edge_case_date_literals
 
-      #if work.genre_string != nil && work.genre_string.include?('Oral histories')
-      #  @implementation = TreatAsOralHistory.new(@work, collection: @collection, parent_work: @parent_work)
-      #elsif treat_as_local_photograph?
-      if treat_as_local_photograph?
+      if treat_as_oral_history?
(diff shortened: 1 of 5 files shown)


app/models/chf/citable_attributes.rb
```diff
@@ -475,5 +482,48 @@ def archive_location
...
+      end
+
+      def date
+        return nil if work.date_of_work.length == 0
+        begin
+          date_recorded = Date.strptime(work.date_of_work.first, '%Y-%m-%d')
+        rescue ArgumentError
+          return nil
+        end
+        date_recorded ? CiteProc::Date.new([date_recorded.year]) : nil
```
jrochkind: i think we'd rather the whole date with month and day for these, not just the year. To more closely match suggested format from catalog. 

I'm not totally sure how you create a CiteProc::Date like that either, might have to look at CiteProc code in that gem.

jrochkind:
Needs tests. No reason not to write em now, if you change the desired output, you'll find the existing tests a nice cushion, and you can even do TDD if you need changes.

app/models/chf/citable_attributes.rb
```diff
@@ -35,9 +35,16 @@ def initialize(work,
       @collection = collection.nil? ? work.in_collection_presenters.first : collection
       @parent_work = parent_work.nil? ? work.parent_work_presenters.first : parent_work
       @edge_case_date_literals = !!edge_case_date_literals
-      @implementation = treat_as_local_photograph? ?
-        TreatAsLocalPhotograph.new(@work, collection: @collection, parent_work: @parent_work) :
-        StandardTreatment.new(@work, collection: @collection, parent_work: @parent_work)
+
+      if work.genre_string != nil && work.genre_string.include?('Oral histories')
```
jrochkind: Extract this into a method to parallel the `treat_as_local_photograph?` line? Maybe `cite_as_oral_history?` ?

app/models/chf/citable_attributes.rb
```diff
@@ -35,9 +35,16 @@ def initialize(work,
...
+        @implementation = TreatAsOralHistory.new(@work, collection: @collection, parent_work: @parent_work)
+      elsif treat_as_local_photograph?
+      #if treat_as_local_photograph?
```
jrochkind: remove commented out code

jrochkind:
ps nice work

app/models/chf/citable_attributes.rb
```diff
@@ -475,5 +487,50 @@ def archive_location
       end
     end
 
+    class TreatAsOralHistory < StandardTreatment
+      def title
+        return work.title if work.interviewee.nil? || work.interviewer.nil?
+        time = work.place_of_interview.nil? ? "" : "at #{work.place_of_interview.first}"
+        place = date.nil? ? "" : "on #{date.iso8601}"
```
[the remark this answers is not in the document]
eddierubeiz: Yup! Let me fix...

jrochkind:
The examples in our catalog have "Oral History Transcript 0279" -- we do have that in our metadata, we can prob get it in the citation somehow. I might mess with it just to mess with it?

PR #1216, opened 2019-01-15
Add "editor" field

### LINKED ISSUE #1215: New dropdown label for 'Creator' field
Label: "Editor" - alphabetical order like existing labels, so after "Contributor."

eddierubeiz:
Fixes #1215 .


jrochkind:
SO many files to edit! nice job

spec/views/curation_concerns/base/show.html.erb_spec.rb
```diff
@@ -16,6 +16,7 @@
       w.artist = ['Boo']
       w.author = ['Cheep Cheep']
       w.addressee = ['Koopa']
+      w.editor = ['Cackletta']
       w.engraver = ['Thwump']
```
jrochkind: This is a spec file, as long as you're adding this to the metadata in the file to be shown... add it to the test that makes sure it is on the page too, like it makes sure the others are on the page?

This may apply to other spec files where you added the metadata but didn't necessarily add the test too. 

If it's HARD to add the test... skip it. But in this file it's just following the exact construction used to check other metadata is on page I think?

jrochkind:
I guess when we add metadata at this point, we should try to do it in concert in the new app?  

Want to make a PR to add it there too?  Which will also get you working on new app more -- if you run into code you don't understand and are curious about, ask!

eddierubeiz:
Make sense!

spec/views/curation_concerns/base/show.html.erb_spec.rb
eddierubeiz: Fixed! Thanks for catching this!

PR #1220, opened 2019-01-29
Export to scihist-digicoll

eddierubeiz:
On the chf-sufia server, run `bundle exec rake chf:export`.

Note:
`ONLY_COLLECTIONS='j6731386g' RAILS_ENV=production bundle exec rake chf:export` will export only that one collection (useful for testing.)

The export will create a batch of json files in `tmp/export`.

Move those files into scihist_digicoll's directory, at: `/tmp/import`.
Then, run `bundle exec rake scihist_digicoll:import`.

### DIFF
--- a/app/exporters/collection_exporter.rb
+++ b/app/exporters/collection_exporter.rb
@@ -1,0 +1,15 @@
+class CollectionExporter < Exporter
+  def edit_hash(h)
+    h['members'] = members
+    h
+  end
+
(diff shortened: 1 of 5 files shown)


jrochkind:
Can you add the instructions from this PR to comments above the rake task? The `ONLY_COLLECTIONS` thing, where the files end up written, etc.

PR #1236, opened 2019-04-02
Ingest audio derivatives; correct derivatives are created on ingest.

eddierubeiz:


jrochkind:
I'm not a huge fan of the architectural choices (inconsistency with how audio is handled compared to other derivs), but fitting into our technical debt in this app is challenging, and we don't want to spend lots of time fixing technical debt in this app we're trying to leave, so I'll say thumbs up anyway. 

We should figure out how to test on travis one way or another, either get ffmpeg installed, or add additional tests that just test the ffmpeg command lines are as expected, maybe with a mock that returns a mock file as output.

PR #1237, opened 2019-04-11
WIP show rudimentary html5 player on front end

eddierubeiz:
- [x] better filenames for mp3 derivative download (in content-disposition header), that use the same logic and match the filenames for image derivatives (eg jpeg)

- [x] "Original file type" output in right-hand metadata column should output proper labels for FLAC and MP3, rather than empty space

### DIFF
--- a/app/helpers/member_helper.rb
+++ b/app/helpers/member_helper.rb
@@ -76,6 +76,7 @@ module MemberHelper
       download_options.each do |option_config|
         list_elements << dropdown_menuitem(
           link_to(option_config[:url],
+            download: '', #forces the download of audio files
             data: {
               content_hook: "dl-original-link",
               analytics_category: "Work",
--- a/app/helpers/member_helper.rb
+++ b/app/helpers/member_helper.rb
@@ -76,7 +76,6 @@ module MemberHelper
       download_options.each do |option_config|
         list_elements << dropdown_menuitem(
           link_to(option_config[:url],
-            download: '', #forces the download of audio files
             data: {
               content_hook: "dl-original-link",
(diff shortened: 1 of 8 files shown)


app/presenters/chf/file_set_presenter.rb
```diff
@@ -2,7 +2,7 @@ module CHF
   # Add some things we can use for our custom UI
   class FileSetPresenter < Sufia::FileSetPresenter
 
-    delegate :original_file_id, to: :solr_document
+    delegate :original_file_id, :original_checksum, to: :solr_document
```
eddierubeiz: This can actually be removed.

app/presenters/chf/file_set_presenter.rb
```diff
@@ -49,5 +49,22 @@ def representative_width
       width
     end
 
+    def audio?
+      representative_content_type =~ /audio/
+    end
+
+    def audio_derivative_links
+      return unless audio?
+      #other_checksum = FileSet.find(id).files.first.checksum.value
```
eddierubeiz: Get rid of these comments

app/helpers/member_helper.rb
```diff
@@ -76,6 +76,7 @@ def member_download_menu(member, parent:, labelled_by: nil, filename_base: nil,
       download_options.each do |option_config|
         list_elements << dropdown_menuitem(
           link_to(option_config[:url],
+            download: '', #forces the download of audio files
```
eddierubeiz: Remove this -- we can modify the URL to force a download -- we already have code for this.

PR #1243, opened 2019-04-30
Provenance metadata

### LINKED ISSUE #1242: New field for provenance
Please create a free-text field entitled "Provenance" to be located underneath "Creator."  

Instructional form text should read as follows:
For fine art materials, input provenance information following Provenance Style Guide document (P:\Eddleman Institute\Museum\Collections\Art Collections\Provenance\Provenance_Style_Guide.docx). Provenance information can either be found in Past Perfect or within provenance files in the Museum department. 

Thank you!

eddierubeiz:
Fixes #1242


eddierubeiz:
I'm reviewing the documents at P:\Eddleman Institute\Museum\Collections\Art Collections\Provenance, which contain a couple references to example provenance statements, including:
https://www.mfa.org/collections/object/double-portrait-of-trudl-33706
https://www.mfa.org/collections/object/figure-of-a-chinese-man-466232
For now I'm proceeding on the assumption that this isn't just a raw text field and we're going to need a bit of HTML -- line breaks at the very least.

jrochkind:
We can possibly apply the same formatting we use for "description" fields?  If line breaks are supposed to be supported, then of course it needs to be an HTML multi-line text entry.

eddierubeiz:
Yup. Just confirmed w/ Hillary. All the same functionality as "Description".

app/controllers/catalog_controller.rb
```diff
@@ -43,13 +43,13 @@ def self.chf_search_fields
...
       "#{solr_name("printer", :stored_searchable)}^500",
-      "#{solr_name("provenance", :stored_searchable)}^500",
       "#{solr_name("publisher", :stored_searchable)}^500",
       "#{solr_name("genre_string", :stored_searchable)}^500",
 
       "#{solr_name("subject", :stored_searchable)}^100",
 
-      "#{solr_name("description", :stored_searchable)}^50",
+      "#{solr_name("description", :stored_searchable)}",
+      "#{solr_name("provenance", :stored_searchable)}^50",
```
jrochkind: you accidentally removed the ^50 from description....

PR #1245, opened 2019-05-03
WIP provenance notes metadata

eddierubeiz:
Tests don't pass yet
No show-hide functionality


jrochkind:
I'm confused what's going on with the "quick and dirty" collapse. I think Boostrap has a built-in collapse/expand JS behavior. It looks like even in Bootstrap 3, which we have here: https://getbootstrap.com/docs/3.4/javascript/#collapse

I think your code may be using it? I'm confused why we need any special hacks (such as custom styles) on top of the built-in BS3 collapse functionality. Thoughts?

jrochkind:
Reading the Medium post you linked to -- it looks like the hack is to make the first couple lines of a paragraph display when "collapsed"?  When seperating "provenance" and "provenance notes" (Is that for sure decided upon? Let's discuss it at our coming up meeting to make sure) -- why can't the "provenance notes" field just be entirely collapsible, using the standard BS functionality?

app/presenters/curation_concerns/generic_work_show_presenter.rb
```diff
@@ -254,6 +254,15 @@ def field_value field, options = {}
...
     end
 
+    # The string NOTES, in all caps, followed by an optional colon,
+    # needs to be on its own line in the Provenance field.
+    # The regular expression also captures extra blank space
+    # before and after the match, as well as before and after the word "NOTES".
+    def split_provenance
+      return [nil, nil] if provenance.nil? || provenance == []
+      return ["",  nil] if provenance == ""
+      provenance.first.split(/\s*\n\n\s*NOTES:?\s*\n\n\s*/m)
```
jrochkind: nice-looking regexp. Did you find "/m" was necessary?  I'm not sure it is?

Also not sure if we need `\r?\n` for every `\n` to get any potential Windows-style newlines that are `\r\n`.

jrochkind:
@eddierubeiz Lookin good. 

I think it would be good to have a test for "What if the data contains TWO `NOTES` markers?"  This is not "good data", but any data that can happen will eventually, and we want to make sure at minimum it doesn't cause an uncaught exception, but best is probably the second one is "ignored" meaning everything after the first one ends up in "notes", it doesn't leave out stuff after the second one. 

I think it will be easier to do that with an additional unit test on the `split_provenance` method itself, instead of just the view test  -- can just test that you still get only two things back with the second including everything even with two `NOTES`. 

The [ruby split method](https://ruby-doc.org/core-2.2.0/String.html#method-i-split) can take a second argument "limit", just passing `2` for it should prob make the hypothetical test pass if it doesn't already.

app/presenters/curation_concerns/generic_work_show_presenter.rb
eddierubeiz: /m is multiline.
Good call re: windows newlines.
I'm leaning towards `/\s*\n\s*NOTES:?\s*\n\s*/m` since `\s` captures `\r`; let me see how that pans out.

jrochkind: I think _maybe_ the only effect of the multiline flag is to have `.` match newlines. This regex doesn't have any `.` in it, let alone ones that are supposed to match newlines. I think maybe it is superfluous?

eddierubeiz:
GOOD CALL re: potential return value having > 2 items.

app/presenters/curation_concerns/generic_work_show_presenter.rb
eddierubeiz: I suspect we do need /m, but I'll check; stay tuned!

jrochkind:
If you're writing "unit" level tests on `split_provenance` itself, which I think probably makes sense, you can also add test cases for the "special cases" in your code, `return [nil, nil] if provenance.nil? || provenance == []` and the other one. Not totally sure the second special cases is needed. One way to find out is write a test case, then anyone can see if it still passes if you remove the special case in the code. :)

jrochkind:
Also in rails, `provenance.blank?` is a good shortcut for `provenance.nil? || provenance == []` when `provenance` might be nil or an array.

app/presenters/curation_concerns/generic_work_show_presenter.rb
eddierubeiz: /m has no effect! You were right!

PR #1246, opened 2019-05-08
Provenance notes: Adjusting data-entry instructions; accepting Notes: as well as NOTES:

eddierubeiz:


app/presenters/curation_concerns/generic_work_show_presenter.rb
```diff
@@ -270,7 +270,7 @@ def split_provenance(prov=provenance)
         raise ArgumentError.new("Provenance should be either nil or an array of length 1.")
       end
       return [nil, nil] if prov.blank? || prov.first.blank?
-      prov.first.split(/\s*\n\s*NOTES:?\s*\n\s*/, 2)
+      prov.first.split(/\s*\n\s*(NOTES|Notes):?\s*\n\s*/, 2)
```
jrochkind: If you add an `i` on the end of the regex instead (`/notes/i`), it will be case-insensitive, and also accept eg "notes" instead of just `NOTES` and `Notes`. 

Although that might not really be an issue, I will go ahead and merge this for now!

PR #1256, opened 2019-05-21
Audio playlist on item display page.

### LINKED ISSUE #1231: Audio file support master ticket
We are going to add support for audio file ingest. 

The [current plan](https://chemheritage.atlassian.net/wiki/spaces/HDCSD/pages/522125313/Time+based+media) is for files to be ingested into Digital Collections app as either FLAC or MP3 (usually FLAC, MP3 for some legacy files with MP3 originals). 

We plan to provide an [HTML5 audio element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/audio) to play files in browser. (At least for first implementation, simplest thing that might work).  

Different browsers support different file formats for playing in browser. According to my research, WebM seems to be the currently preferred format supported by modern browsers. That plus MP3 seems to hit almost any browser still in use. 
* Note that for FLAC originals, either of these would be a derivative. 
* Even if original was MP3, in the sufia architecture where originals are delivered by the Rails app proxying fedora -- I'm not sure this will provide a good experience as an HTML5 audio source, without bringing down our apps. Using derivatives which are delivered directly from S3 is safer all around. 
* We may choose to deliver just WebM to hit almost everyone while saving the extra space and time of also having an MP3, or we may have an MP3 too. Need to do more research and consult with stakeholders, exact derivatives we need are not yet fixed. 
* Later we may move to [HLS format](https://developer.mozilla.org/en-US/docs/Web/Guide/Audio_and_video_delivery/Setting_up_adaptive_streaming_media_sources#HLS_Encoding).   But that's somewhat more complicated, and not natively supported by many browsers (can be played with JS support like [mediaelement.js](https://www.mediaelementjs.com/) and possibly [HLS.js](https://github.com/video-dev/hls.js/), which I believe Avalon uses both. (HLS is probably more important/beneficial for video than audio. Many/most people supporting HLS also use a "media server" to simplify serving HLS, serving HLS is a lot more complex).  For now, we're going to start with HTML5 and the simpler audio formats it supports. 
  * (or dash format? don't know much about that)

We also need to decide if we want to offer additional derivative types for download. Like offer someone a download option of an MP3 version even if the original was FLAC? If we're going to make and store an MP3 as a derivative for dl anyway, obviously might as well supply it as an alternate format to html5 audio tag. 

## Order of implementation

1. Get it so you can add a FLAC or MP3 audio file to the Sufia app at all. 
  * Right now trying to upload one will possibly result in errors, as ingest-related code that assumes everything is an image might try to treat the audio file as an image, and raise.  
  * Make sure that doesn't happen. Start with _no_ derivatives being created for audio files. 
  * This step is done when you can complete the steps to ingest a file with an audio file, and then go look at that audio file in the front-end UI, all without getting any 500s. You can "download original" for audio file, and get the original.  At this step, in the space for a thumbnail, it may be blank or a missing image placeholder, that's fine or expected. 

2. Add derivatives  creation for audio files. 
  * Start with WebM. Maybe MP3 too. 
  * Not sure the best way to convert a FLAC or MP3 to these other elements. Perhaps [ffmpeg command line](https://ffmpeg.org/ffmpeg.html#Video-and-Audio-file-format-conversion)?  May need to make sure ansible is installing ffmpeg on deploy machines, if it's not already. 
  * We need to consider appropriate bitrate and compression level for derivatives. Assuming human conversation rather than music? We want to minimize file size without noticeable drop in perceived quality.  64 bps may be enough. For MP3s, "V2" compression level is probably good, equivalent in WebM-ogg. 
  * This step is done when the derivatives have been created and _exist_, they aren't yet showing up in any UI yet. 

3.Now we need an actual UI that allows playability and downloadability
* Let's eddie and jonathan talk before implementing. But basically. 
* For first run-through, we're not going to change where things are on the page, but the component on the page that displays a "thumbnail" has to be changed to deal with audio.  But there's still one component on the page per fileset, with functionality only about that fileset. 
* On search results page, it should probably display a generic "audio" icon (maybe use SVG). 
* On item detail page, it should display an HTML5 audio player for an audio file in the current "thumbnail" spot. (Initially, don't worry about the fact that it's gonna be too small in the existing CSS for the thumbnail spot, just get it on there.) It should also display the staff-editable title/label.
* It should still have a download options menu, but display appropriate download options for audio. The original, and whatever derivatives we decide we want to offer for download. 

4. We'll presumably have another round of front-end UX tweaks after we have a basic demo as per 3 above, premature to spec that now. 

1-3 should be done as separate PR's. (or even more individual PR's than that is okay too). 

## New App

We need to do these things in parallel in New App. At least we need to do 1 and 2 above, we don't have a 3 in new app yet really. @eddierubeiz I'd like to propose trying out a workflow this time where you do 1 in old app, then do 1 in new app. Then 2 in old app, then 2 in new app. Instead of 'finishing' old app first then moving to new app. This will ensure that we don't have the option of neglecting new app, the feature is not done until it's in both. And should help minimize "context switching" cost.

eddierubeiz:
Will close #1231

* Show an audio playlist above the representative display -- practically speaking at the top left of the page, or at the very top on narrow screens.
* If you are logged in, you will also see the regular audio tags below the main item display.
 * Otherwise, they will be hidden; your only way to interact with audio will be through this playlist.
* Show the download links only for the selected track, per Hillary.

**TODO:**
- [x] Get rid of the grey background around the audio tag.
- [x] The audio tag should be as wide as possible.
- [x] Rename viewable_audio_member_presenters to a more helpful method name
- [x] Merge _media_player.html.erb and _media_item.html.erb .
- [x] show_thumb_member_presenters should filter out audio unless the user is staff.
- [x] Investigate mysterious and probably unnecessary changes to _show_page_image.html.
- [x] Move the CSS to a new playlist.scss file in app/assets/stylesheets/local/, where it will get picked up by  `@import "local/**/*";` in sufia.scss
- [x] Convert track list table to an actual HTML list
- [x] Edit front-end tests
- [x] Edit CSS to work well on narrow phone display
- [x] Automatically play the next track when you get to the end of a track


app/assets/stylesheets/application.css
```diff
@@ -16,9 +16,14 @@
...
- 
+ *
  *
  * Used by blacklight_range_limit
  *= require  'blacklight_range_limit'
  *
+ * TEMOPORARY -- fold this into work-show.scss before merging.
+ * Audio player:
+ *= require playlist.css
+ *=
```
jrochkind: It should be in `stylesheets/local/playlist.scss` (note change suffix to `scss`, why not), and then it doens't need a `require`, it will get automatically picked up by our scss `@import` of everything in local (which happens to be in our local `sufia.scss` for confusing legacy reasons).

app/assets/stylesheets/local/work-show.scss
```diff
@@ -216,6 +216,16 @@
...
+  .show-page-audio-playlist-wrapper {
+    vertical-align: bottom;
+    margin: $thumb-horizontal-margin $thumb-horizontal-margin ($line-height-computed * 1.5) 0;
+    .btn {
+      @extend .btn-sm;
+      border-radius: 2px;
+    }
+  }
+
+
```
jrochkind: maybe move this to playlist.css?

app/presenters/curation_concerns/generic_work_show_presenter.rb
```diff
@@ -178,6 +178,11 @@ def direct_representative_presenter
       @direct_representative_presenter ||= member_presenters([representative_id]).first
     end
 
+    # Used by _media_player.
+    def viewable_audio_member_presenters
+      viewable_member_presenters.select {|x| x.audio?}
+    end
```
jrochkind: maybe a better name for this helper method that makes it clear this is a list of things that will be used with the audio player, displayed in a menu, when you click on them they will play in a player.

app/views/curation_concerns/base/_media_item.html.erb
```diff
@@ -0,0 +1,9 @@
+<tr class="track_listing"
+    data-title="<%= member.title.first %>"
+    data-member-id="<%= member.id %>"
+    data-mp3-url="<%= CHF::AudioDerivativeMaker.s3_url(file_set_id:member.id, file_checksum:member.representative_checksum, type_key: :standard_mp3) %>"
+    data-webm-url="<%= CHF::AudioDerivativeMaker.s3_url(file_set_id:member.id, file_checksum:member.representative_checksum, type_key: :standard_webm) %>"
+>
+    <!-- TODO instead of a table, make this a list of links.  -->
+    <td class="title" title="Listen to '<%= member.title.first %>' "><%= member_counter + 1 %>. <%= member.title.first %></td>
```
jrochkind: merge this into _media_player.html.erb, don't think it needs to be a separate template, although I see why you did it it's not totally unreasonable.

app/views/curation_concerns/base/_show_page_image.html.erb
```diff
@@ -9,71 +9,78 @@
       * extra_classes: optional. extra CSS calsses added to the wrapper <div> tag.
 
 %>
-<% wrapper_tag = member.representative_content_type =~ /audio/ ? 'show-page-audio-wrapper' : 'show-page-image-wrapper' %>
+<%
+  # Audio content in *this* list is only shown to staff.
+  # However, *all* users can access an audio player which is displayed above the hero image.
+  if current_ability.current_user.staff? || (member.representative_content_type =~ /audio/).nil?
+%>
```
jrochkind: we do not need staff-specific logic in the template, rather we should edit the `show_thumb_member_presenters` in a presenter, so it does not include audio unless the user is staff, there.

jrochkind: _show_page_image.html may not have to be changed at all in this PR? Or it can have a lot less.

jrochkind:
We also may have to change the filename that the browser saves the file as when you download? That's set via an instruction to S3 when creating a signed link, instructing S3 to set a "Content-Disposition" header. 

Not sure if it's wanted to be the "original" filename, or the "label" that Nicole et al will edit in the admin interface? 

I recall that filename of the downloaded file was important to OH, is why I bring it up for attention. 

Right now, it uses the same logic as all our downloads, which is first 3 "words" of the file title/label, plus the work ID ("digital collections internal ID" eg `rj4305516`) for uniqueness.

jrochkind:
@eddierubeiz I think I am getting the audio player showing up on every work page, not just ones with audio? 

Hmm, maybe not every one. But I created a new Work in my dev instance with _no_ files attached... and it's got a player! Which is labelled "Current track", and can't actually play anything. So this seems like a bug.

jrochkind:
Not always followed in our codebase, but using hyphens instead of underscores in CSS classes is the standard. `track-listing` instead of `track_listing`. 

Also, instead of creating a new `$thumb-horizontal-margin`, would be better to re-use an existing bootstrap spacing variable of some kind, possibly a portion of it like `($some-existing-variable / 2)`. I'll see what I find.

jrochkind:
When I change a class from `now_playing_info` to `now-playing-info`, I break the JS, since it used class names. 

Best practice is for JS to trigger on `data` attributes rather than class names. 

Also I don't like that "placeholder" test like `Current track.` and `src="mp3_url "` -- if those should never actually be shown to the user, they should be left blank. `src="mp3_url"` is never gonna work, so better to leave it blank than put placeholder that isn't going to work there.

jrochkind:
@eddierubeiz the JS and CSS file are called `playlist`, but the partial is called `media_player`. Do you think we should make em match?

Also, can you look at other outstanding unchecked checkboxes, and remove ones no longer relevant, attend to ones still relevant? I'm not sure what they all mean or which ones have been done. Thanks!

app/views/curation_concerns/base/_media_item.html.erb
eddierubeiz: Yup, I folded these in.

app/views/curation_concerns/base/_media_player.html.erb
```diff
@@ -0,0 +1,51 @@
+<% if @presenter.audio_playlist_presenters.present? %>
```
jrochkind: I would maybe put this `if` into the wrapper that calls this instead. 

    <% if @presenter.audio_playlist_presenters.present? %>
      <%= render 'media_player', presenter: @presenter %>
    <% end %>

But I could be wrong or it might not matter and be no better, so I won't hold up the merge any longer!

jrochkind: Oh no, but writing that comment alerted me to something else. 

You are passing in a **local** param to `media_presenter` template. That's what `render 'media_player', presenter: @presenter`. Take the iVar `@presenter`, pass it in a local variable `presenter`. In general iVars are hard to keep track of and end up tightly coupling the partial template to the controller -- since iVars are set in the controller. So you gotta know what the controller is doing (which could be many levels up) to know if the iVars are set right. 

So it is considered better practice to pass in a local -- now to understand the context of the partial template, you only you only have to look at where it's called and what locals are passed in.  Use iVars in top-level templates, but locals in partials. (This is one of the messy parts of Rails). 

But to use a local `presenter` passed into a template, it's just in the local variable `presenter`. 

In this template, you're still writing `@presenter`. So you're passing a local argument in... but then not using it, instead using the iVar `@presenter` set by the controller. 

It should be one or the other not both. Probably switch to using local argument `presenter` in this template.

PR #1272, opened 2019-07-08
Added list of SEO or Ad bots to block to robots file

### LINKED ISSUE #1271: Block common bots not associated with search engines
We should start blocking bots whose scraping of the site doesn't actually add our content to search engines or other tools that help people find us.
Current bots to block
- MauiBot
- centurybot9@gmail.com
- AhrefsBot
- SemrushBot
- BLEXBot

The first two are somewhat mysterious as to their purpose and owners, the last 3 are well described and clearly note they follow robot.txt and the limits they work under which are not unreasonable. Since we don't use or want ads or external marketing on our site though, blocking them should save us some traffic and reduce some of the issues we've had hopefully.

sanfordd:
Closes #1271

This may take a few days from being pushed out until we see any notable changes.

### DIFF
--- a/app/views/application/robots.txt.erb
+++ b/app/views/application/robots.txt.erb
@@ -12,6 +12,18 @@ User-agent: *
 Disallow: /downloads/
 Crawl-delay: 5
 
+User-agent: MauiBot 
+Disallow: /
+
+User-agent: BLEXBot
+Disallow: /
+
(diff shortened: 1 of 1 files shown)


jrochkind:
Thanks @sanfordd ! Will merge, and deploy chf_sufia prodution with it.

PR #1285, opened 2019-08-20
Test a few more things on oai_pmh feature spec

jrochkind:
To have a somewhat more complete spec to copy over to new app, while verifying here that old app meets spec.


jrochkind:
Failing for unrelated reasons, not sure why.

PR #1287, opened 2019-11-05
new url for sufia application

sanfordd:
This'll let us deploy to the correct server since we've switched the old URL to the new application.


jrochkind:
Huh, not sure what has gotten messed up in travis, I'll look into it.

PR #1289, opened 2019-11-06
Tell travis to update rubygems before running

jrochkind:
Try get around weird rubygems bug that's suddenly started affecting us:
https://bundler.io/blog/2019/05/14/solutions-for-cant-find-gem-bundler-with-executable-bundle.html

Looks like this in travis log with test failure:

```
/home/travis/.rvm/rubies/ruby-2.5.1/lib/ruby/2.5.0/rubygems.rb:289:in `find_spec_for_exe': can't find gem bundler (>= 0.a) with executable bundle (Gem::GemNotFoundException)
```

Not sure why it doesn't reproduce locally, just on travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+dist: trusty
 language: ruby
 script: bundle exec rake dev:spec_with_app_load
 rvm:
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,10 @@ addons:
     - ffmpeg
 # END NOTE
 
-
+before_install:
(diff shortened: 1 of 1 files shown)


jrochkind:
I think real issue may be this one:

https://travis-ci.community/t/error-installing-oraclejdk8-expected-feature-release-number-in-range-of-9-to-14-but-got-8/3766/8

Travis changed the underlying platform/OS setup you are running on, which made some things stop working. But putting `dist: trusty` in may make it use the same one that was working before again, and make it work again. Let's see.

PR #1292, opened 2019-11-06
Date export script

eddierubeiz:
We do not currently have any plans to change any of the dates in the new digital collections ap to bring the created-date and modified-date sorts closer in line with Sufia.

However, this script outputs out the relevant dates so that, if we change our minds, we have them ready to go. (It's much easier to get at them now, while the old app is still running, than extracting them from the fedora export.)

### DIFF
--- a/app/exporters/dump_sufia_dates.rb
+++ b/app/exporters/dump_sufia_dates.rb
@@ -1,0 +1,38 @@
+# This produces a (valid) json dump of Collection, GenericWork
+# and FileSet dates.
+
+# bin/rails  runner dump_sufia_dates.rb
+# cat  sufia_dates.json   |  python -m json.tool
+def to_utc(some_date)
+  return nil if some_date.nil?
+  return some_date.utc if some_date.class == DateTime
+  return DateTime.rfc2822(some_date).utc if some_date.class == String
+end
+
+progress_bar_total = FileSet.count + GenericWork.count + Collection.count
+progress_bar = ProgressBar.create(total: progress_bar_total, format: "%a %t: |%B| %R/s %c/%u %p%% %e")
+
+open('sufia_dates.json', 'w') do |f|
+  f.puts '{"date_list":['
+  comma = false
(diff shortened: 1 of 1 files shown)


app/exporters/dump_sufia_dates.rb
```diff
@@ -0,0 +1,38 @@
...
+      f.puts "," if comma
+      comma = true
+      tempHash = {
+        date_modified: to_utc(x.date_modified),
+        date_uploaded: to_utc(x.date_uploaded),
+        create_date: to_utc(x.create_date.utc),
+        type: x.class.to_s,
+        id: x.id
+      }
+      f.puts tempHash.to_json
```
jrochkind: If this is meant to be parseable as json, I think you need to put a comma after each tempHash.to_json...

Alternately you could skip the surrounding `{"date_list":['` stuff, and just call it ["newline-delimited json".](http://ndjson.org/) -- if you make sure `tempHash.to_json` is a single line with no newlines!

Or maybe I'm wrong (this is just from a code-read, I didn't try to parse it), and everything is fine?

Let's make sure we've tested we can actually read this stuff before merging.  Could be manual (maybe you've already done it and can just tell me), or we could figure out a way to unit test it somehow.

eddierubeiz: Hey -- the json parses fine:
```
bin/rails  runner dump_sufia_dates.rb
cat  sufia_dates.json   |  python -m json.tool
```
will parse and output a prettified version of it.

jrochkind: OK, can't figure out how that's working, either there are commas I wans't aware of, or the python tool can handle the fact that they are illegally missing, but either way I guess that works, and is documented here as a way to turn it into good json.

jrochkind:
@eddierubeiz says:

> Hey -- the json parses fine:

>      bin/rails  runner dump_sufia_dates.rb
>      cat  sufia_dates.json   |  python -m json.tool
> will parse and output a prettified version of it.

